PR #2, opened 2014-02-03
ts refactor #0

keroserene:
With a focus on the "client" TCP and SOCKs implementation and a tcp echo server.
End-to-End works.


src/client/tcp.ts
```diff
@@ -0,0 +1,457 @@
...
+    https://github.com/GoogleChrome/chrome-app-samples/tree/master/tcpserver
+*/
+
+/**
+ * Converts an array buffer to a string of hex codes and interpretations as
+ * a char code.
+ *
+ * @param {ArrayBuffer} buf The buffer to convert.
+ */
+function getHexStringOfArrayBuffer(buf) {
```
iislucas: We probably don't want these functions in public scope; move into TCP module? Or move to where it's used.

src/client/tcp.ts
```diff
@@ -0,0 +1,457 @@
...
+
+  /**
+   * Converts an array buffer to a string
+   *
+   * @private
+   * @param {ArrayBuffer} buf The buffer to convert.
+   * @param {Function} callback The function to call when conversion is
+   * complete.
+   */
+  function _arrayBufferToString(buf, callback) {
```
iislucas, line 416: Put underscore at the end for JS style.

iislucas:
LGTM, with minor tweaks to names; feel free to make TODOs.

PR #14, opened 2014-02-06
Sword2

quantsword:
Somehow the change in socket_chrome was lost. This change add that back. It also changed the way how onData is handled. Instead of handling it in socket level, now it is done in TCP server level.

### DIFF
--- a/chrome/lib/socket_chrome.js
+++ b/chrome/lib/socket_chrome.js
@@ -48,7 +48,7 @@ var readSocket = function(socketId) {
       this.dispatchEvent('onData', arg);
       chrome.socket.read(socketId, null, dataRead);
     } else if (readInfo.resultCode === 0) {
-      console.error('socket ' + socketId + ' has been closed.')
+      console.log('socket ' + socketId + ' has been closed.')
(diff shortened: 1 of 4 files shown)


chrome/lib/socket_chrome.js
```diff
@@ -26,30 +46,25 @@ var readSocket = function(socketId) {
...
-      // can happen in usual program flow, so we will not log the error.
-      // console.warn('Got a disconnection for socket ' + socketId);
-      if (readInfo.resultCode === -2) {
-        console.log('HACKITY HACK: Ignoring an unexpected -2 from a socket.  ' +
-            'Deal with it later');
-      }
-      this.dispatchEvent('onDisconnect', {socketId: socketId});
+      chrome.socket.read(socketId, null, dataRead);
+    } else if (readInfo.resultCode === 0) {
+      console.error('socket ' + socketId + ' has been closed.')
```
keroserene: This shouldn't be an error, as it's normal behavior? (otherwise the log looks really scary)

Thankfully, it seems the "unexpected socket" bug from before isn't happening anymore when I build this version :)

PR #15, opened 2014-02-06
git typescript interface for promises

iislucas:


keroserene:
Cool! can we just call the directory 'interfaces', since ts is implied when src/ only contains typescript?

src/interfaces/promise.d.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Generic Promise class.
+ *
+ * F is the `fullfilled` type given to onFulfilled function.
+ * R is the `rejected` type given to the onRejected object.
+ */
+declare class Promise<F,R> {
+
+  constructor (onFulfilled:(fulfillObj : F) => void,
+               onRejected:(rejectObj? : R) => void);
```
keroserene, line 17: Actually, there seems to be a subtle difference in the way js promises are constructed.
Instead of:
`new Promise(onFulfilled, onRejected)`
It's actually:
`new Promise(function(onFulfilled, onRejected) { ... use the fulfil or rejection callbacks as necessary })`

So I'll update this when we're actually using the promises.

iislucas, line 17: nice catch! thanks!

On Fri, Feb 7, 2014 at 3:47 PM, Serene H notifications@github.com wrote:

> In src/interfaces/promise.d.ts:
> 
> > +
> > +interface Thenable<F,R> {
> > -  then:(onFulfilled:(F) => void,
> > -        onRejected:(R) => void) => Thenable<F,R>;
> >   +}
> >   +
> >   +/**
> > - \* Generic Promise class.
> > - *
> > - \* F is the `fullfilled` type given to onFulfilled function.
> > - \* R is the `rejected` type given to the onRejected object.
> > - */
> >   +declare class Promise<F,R> {
> >   +
> > -  constructor (onFulfilled:(fulfillObj : F) => void,
> > -               onRejected:(rejectObj? : R) => void);
> 
> Actually, there seems to be a subtle difference in the way js promises are
> constructed.
> Instead of:
> new Promise(onFulfilled, onRejected)
> It's actually:
> new Promise(function(onFulfilled, onRejected) { ... use the fulfil or
> rejection callbacks as necessary })
> 
> So I'll update this when we're actually using the promises.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uProxy/socks-rtc/pull/15/files#r9554529
> .

## 

Lucas Dixon | Google Ideas

PR #17, opened 2014-02-07
ts refactor #3

keroserene:
- Chrome freedom sockets typescripted.
- Begin 'src/interfaces/' directory with .d.ts files, such as for PeerConnection
- Cleaned and documented much more

TESTED: manual end-to-end


src/chrome-fsocket.ts
```diff
@@ -0,0 +1,126 @@
+/**
+ * Chrome sockets over freedom sockets.
+ * TODO: This should be refactored into freedom someday...
+ */
+
+declare var chrome:any;
+
+
+// Platform independent interface. TODO: Put somewhere so firefox can use it.
```
iislucas, line 9: add TODO: provide fully detailed types, not just "any". This should end up living in something like freedom-socket.d.ts

src/rtc-to-net/netclient.ts
```diff
@@ -20,90 +21,98 @@ module Net {
...
-  */
+  /**
+   * Net.Client - TCP socket connection to a net destination.
+   *
+   * TODO: write a unit test using this and tcp-server.
+   */
   export class Client {
 
     socketId:string = null;
     queue:any[] = [];
```
iislucas, line 32: add TODO: more specific typing on queue.

iislucas:
LGTM. Nice! Eventually we'll get rid of all the "any" types...

PR #18, opened 2014-02-07
fix ipv6 host interpretation.

quantsword:


iislucas:
Nice! How was it tested? (end-to-end by hand?)

quantsword:
Yes, this is manually tested using curl. Curl offer way to control if dns should be resolved remotely.  The 2nd version in following did dns resolution in remote. So to test this, you will need the first version. www.google.com usually returns ipv6 address for me, but not always. 
curl --socks5 localhost:9999 <site>  
curl --socks5-host localhost:9999 <site> 

Serene has plan to provide some end-to-end test case right? Probably we should include such a case once the testing framework is established.

PR #21, opened 2014-02-10
promises - socks-to-rtc side

keroserene:
Convert most callbacks floating between SOCKS server and TCP server to promises.
- TCP server listening, accept, disconnect
- TCP connection create and disconnect
- SOCKs handshake and request handling
- Slightly cleaner documentation

Tested: manual end-to-end

### DIFF
--- a/src/interfaces/promise.d.ts
+++ b/src/interfaces/promise.d.ts
@@ -13,8 +13,8 @@ interface Thenable<F,R> {
  */
 declare class Promise<F,R> {
 
-  constructor (onFulfilled:(fulfillObj : F) => void,
-               onRejected:(rejectObj? : R) => void);
+  constructor (f:(onFulfilled:(fulfillObj:F)=>void,
+                  onRejected:(rejectObj?:R)=>void)=>void);
 
   then<F2,R2> (onFulfilled:(F) => Promise<F2,R2>,
                onRejected?:(R) => void)
--- a/src/interfaces/promise.d.ts
+++ b/src/interfaces/promise.d.ts
@@ -1,47 +1,48 @@
 // Based on http://www.html5rocks.com/en/tutorials/es6/promises/#toc-api
 
-interface Thenable<F,R> {
-  then:(onFulfilled:(F) => void,
-        onRejected:(R) => void) => Thenable<F,R>;
+interface Thenable<T> {
+  then:(fulfill:(t:T) => void,
+        reject?:(e:Error) => void) => Thenable<T>;
 }
 
 /**
- * Generic Promise class.
+ * Generic Promise for built-in js Promises.
  *
- * F is the `fullfilled` type given to onFulfilled function.
- * R is the `rejected` type given to the onRejected object.
+ * T is the `fullfillment object` type given to onTulfilled function.
+ *
+ * The rejection object is always a javascript Error.
  */
-declare class Promise<F,R> {
+declare class Promise<T> {
 
-  constructor (f:(onFulfilled:(fulfillObj:F)=>void,
-                  onRejected:(rejectObj?:R)=>void)=>void);
+  constructor (f:(fulfill:(t:T)=>void,
+                  reject:(e:Error)=>void)=>void);
 
-  // |onFulfilled| either returns a promise...
-  then<F2,R2> (onFulfilled?:(F) => Promise<F2,R2>,
-               onRejected?:(R) => void)
-      :Promise<F2,R2>;
+  // |onTulfilled| either returns a promise...
+  then<T2> (fulfill:(t:T) => Promise<T2>,
+            reject?:(e:Error) => void)
+      :Promise<T2>;
 
   // or the next fulfillment object directly.
-  then<F2,R2> (onFulfilled?:(F) => F2,
-               onRejected?:(R) => void)
-      :Promise<F2,R2>;
+  then<T2> (fulfill?:(t:T) => T2,
+            reject?:(e:Error) => void)
+      :Promise<T2>;
 
-  catch (catchFn:(rejectObj:R) => void)
-      :Promise<F,R>;
+  catch (catchTn:(e:Error) => void)
+      :Promise<T>;
 
-  static resolve<F,R> (thenable:Thenable<F,R>)
-      :Promise<F,R>;
+  static resolve<T> (thenable:Thenable<T>)
+      :Promise<T>;
 
-  static resolve<F,R> (fulfillObj:F)
-      :Promise<F,R>;
+  static resolve<T> (t?:T)
+      :Promise<T>;
 
-  static reject<F,R>  (rejectObj:R)
-      :Promise<F,R>;
+  static reject<T> (e:Error)
+      :Promise<T>;
 
-  static all<F,R> (...args:Thenable<F,R>[])
-      :Promise<F,R>;
+  static all<T> (...args:Thenable<T>[])
+      :Promise<T>;
 
-  static race<F,R> (...args:Thenable<F,R>[])
-      :Promise<F,R>;
+  static race<T> (...args:Thenable<T>[])
+      :Promise<T>;
 
 }
--- a/src/interfaces/promise.d.ts
+++ b/src/interfaces/promise.d.ts
@@ -1,7 +1,7 @@
 // Based on http://www.html5rocks.com/en/tutorials/es6/promises/#toc-api
 
 interface Thenable<T> {
-  then:(fulfill:(t:T) => void,
+  then:(fulfill:(t?:T) => void,
         reject?:(e:Error) => void) => Thenable<T>;
 }
 
@@ -14,7 +14,7 @@ interface Thenable<T> {
  */
 declare class Promise<T> {
 
-  constructor (f:(fulfill:(t:T)=>void,
+  constructor (f:(fulfill:(t?:T)=>void,
                   reject:(e:Error)=>void)=>void);
 
   // |onTulfilled| either returns a promise...
--- a/src/interfaces/promise.d.ts
+++ b/src/interfaces/promise.d.ts
@@ -1,4 +1,5 @@
 // Based on http://www.html5rocks.com/en/tutorials/es6/promises/#toc-api
+// Promise Spec: http://promises-aplus.github.io/promises-spec/
 
 interface Thenable<T> {
   then:(fulfill:(t?:T) => void,
@@ -46,3 +47,10 @@ declare class Promise<T> {
       :Promise<T>;
 
 }
+
+
+// Add .stack attribute to Errors (which actually does exist in js).
+interface Error {
+  stack?:any;
+}
+
--- a/src/interfaces/promise.d.ts
+++ b/src/interfaces/promise.d.ts
@@ -16,7 +16,13 @@ declare class Promise<F,R> {
   constructor (f:(onFulfilled:(fulfillObj:F)=>void,
                   onRejected:(rejectObj?:R)=>void)=>void);
 
-  then<F2,R2> (onFulfilled:(F) => Promise<F2,R2>,
+  // |onFulfilled| either returns a promise...
+  then<F2,R2> (onFulfilled?:(F) => Promise<F2,R2>,
+               onRejected?:(R) => void)
+      :Promise<F2,R2>;
+
+  // or the next fulfillment object directly.
+  then<F2,R2> (onFulfilled?:(F) => F2,
                onRejected?:(R) => void)
       :Promise<F2,R2>;
 
--- a/src/interfaces/promise.d.ts
+++ b/src/interfaces/promise.d.ts
@@ -18,18 +18,19 @@ declare class Promise<T> {
   constructor (f:(fulfill:(t?:T)=>void,
                   reject:(e:Error)=>void)=>void);
 
-  // |onTulfilled| either returns a promise...
+  // |onFulfilled| either returns a promise...
   then<T2> (fulfill:(t:T) => Promise<T2>,
-            reject?:(e:Error) => void)
+            reject?:(e:Error) => Promise<T2>)
       :Promise<T2>;
 
   // or the next fulfillment object directly.
   then<T2> (fulfill?:(t:T) => T2,
-            reject?:(e:Error) => void)
+            reject?:(e:Error) => T2)
       :Promise<T2>;
 
-  catch (catchFn:(e:Error) => void)
-      :Promise<T>;
+  catch (catchFn:(e:Error) => Promise<T>):Promise<T>;
+  catch (catchFn:(e:Error) => T):Promise<T>;
+  catch (catchFn:(e:Error) => void):Promise<void>;
 
   static resolve<T> (thenable:Thenable<T>)
       :Promise<T>;
--- a/src/interfaces/promise.d.ts
+++ b/src/interfaces/promise.d.ts
@@ -28,14 +28,17 @@ declare class Promise<T> {
             reject?:(e:Error) => void)
       :Promise<T2>;
 
-  catch (catchTn:(e:Error) => void)
+  catch (catchFn:(e:Error) => void)
       :Promise<T>;
 
   static resolve<T> (thenable:Thenable<T>)
       :Promise<T>;
 
-  static resolve<T> (t?:T)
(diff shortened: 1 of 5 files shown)


src/interfaces/promise.d.ts
```diff
@@ -1,41 +1,56 @@
...
-               onRejected:(rejectObj? : R) => void);
+  // or the next fulfillment object directly.
+  then<T2> (fulfill?:(t:T) => T2,
+            reject?:(e:Error) => void)
+      :Promise<T2>;
 
-  then<F2,R2> (onFulfilled:(F) => Promise<F2,R2>,
-               onRejected?:(R) => void)
-      :Promise<F2,R2>;
+  catch (catchTn:(e:Error) => void)
```
iislucas: catchTn => catchFn ?

src/interfaces/promise.d.ts
```diff
@@ -1,41 +1,56 @@
...
+      :Promise<T>;
 
-  catch (catchFn:(rejectObj:R) => void)
-      :Promise<F,R>;
+  static resolve<T> (thenable:Thenable<T>)
+      :Promise<T>;
 
-  static resolve<F,R> (thenable:Thenable<F,R>)
-      :Promise<F,R>;
+  static resolve<T> (t?:T)
```
iislucas: I think we'd be best with `resolve<T> (t : T)` and `resolve<void> ()`. That way the type-checker will fix bad uses of void promises.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -38,17 +38,17 @@ module SocksToRTC {
     public start = (options) => {
       console.log('Client: on(start)... ' + JSON.stringify(options));
       // Bind peerID to scope so promise can work.
-      var peerId = this.peerId = options.peerId;
+      var peerId = options.peerId;
       if (!peerId) {
         console.error('SocksToRTC.Peer: No Peer ID provided! Cannot connect.');
         return false;
       }
       this.shutdown();  // Reset everything.
```
iislucas: I don't like shutdown here... let's call it "reset()" if that's what it does?

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -123,7 +123,11 @@ module SocksToRTC {
       // When the TCP-connection receives data, send to sctp peer.
       // When it disconnects, clear the |channelLabel|.
       session.onRecv((buf) => { this.sendToPeer_(channelLabel, buf); });
-      session.onDisconnect(() => { this.closeConnection_(channelLabel); });
+      session.onDisconnect()
```
iislucas: renaming: onDisconnect => onceDisconnected

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -132,18 +136,18 @@ module SocksToRTC {
...
-      connectedCallback({ ipAddrString: '127.0.0.1', port: 0 });
+      return { ipAddrString: '127.0.0.1', port: 0 };
     }
 
     /**
      * Close a particular tcp-connection and data channel pair.
      */
     private closeConnection_ = (channelLabel:string) => {
+      console.log('CLOSE CHANNEL ' + channelLabel);
       if (this.socksSessions[channelLabel]) {
```
iislucas: Can we have a null entry in "this.socksSessions[channelLabel]"? If so, it's strange... no? (I know you didn't write this... so you can fix now or add a TODO :-)  ... I just added a section to our style doc: https://docs.google.com/a/google.com/document/d/1ZV70oTNWO00bs0RcMwqr3FvKJXA6smYQ2kZ8Zr8YZnw/edit#heading=h.d0uggwuleeao

I'd also be inclined to do a check if it's not in the sessions, and if not then raise an error. Unless we know this gets called legitimately for non-existent channel labels. If so, we should comment on that.

src/socks-to-rtc/socks.ts
```diff
@@ -144,161 +146,167 @@ module Socks {
...
+          .then(Socks.Session.CheckVersion)
+          .then(Socks.Session.CheckAuth)
+          // Success - Create new session.
+          .then(() => {
+            conn.on('recv', null);  // Disable recv until session is ready.
+            return Socks.Session.Create(conn);
+          // AUTH error. (Required method not available).
+          }, (e) => {
+            replyToTCP(conn, Socks.AUTH.NONE);
+            return e;
```
iislucas: shouldn't this be `throw(e)` ? When you return something from a catch/error handler, that becomes the new promise value. This makes me think we should actually fix our "then" typing too.

src/socks-to-rtc/socks.ts
```diff
@@ -144,161 +146,167 @@ module Socks {
...
+            console.error('SOCKS Server: ' + e.message);
+            conn.disconnect();
+          });
+    }
 
-      this.tcpServer = tcpServer;
+    listen() {
+      return this.tcpServer.listen().then(() => {
+        console.log('LISTENING ' + this.tcpServer.addr + ':' + this.tcpServer.port);
+        return null;
```
iislucas: What's the return value for?

src/socks-to-rtc/socks.ts
```diff
@@ -144,161 +146,167 @@ module Socks {
...
 
 
   /**
-   * Socks.Session
+   * Socks.Session - layer dealing with handshakes and requests over TCP.
    */
   export class Session {
 
-    private request:any = null;
+    public static GetHandshake(handshake) { return new Uint8Array(handshake); }
```
iislucas: GetHandshake -> getHandshake

src/socks-to-rtc/socks.ts
```diff
@@ -144,161 +146,167 @@ module Socks {
...
-      // We are no longer at waiting for a proxy request on this tcp connection.
-      // TODO: determine if this is necessary.
-      this.tcpConnection.on('recv', null);
-
-      var byteArray = new Uint8Array(buffer);
-      var socksVersion = byteArray[0];
+    /**
+     * Only SOCKS Version 5 is supported.
+     */
+    public static CheckVersion(handshakeBytes:Uint8Array) {
```
iislucas: CheckVersion -> checkVersion

Generally, function names should start with lowercase. Fix in other places too.

http://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml?showone=Naming#Naming

src/socks-to-rtc/socks.ts
```diff
@@ -144,161 +146,167 @@ module Socks {
...
-    public onDisconnect = (callback:(buf)=>void) => {
-      this.tcpConnection.on('disconnect', callback);
-    }
-
-    public disconnect = () => {
-      // TODO: This will fire the disconnect handler from the underlying TCP
-      // connection, but it's actually just some other callback. Clean this up
-      // with promises.
-      this.tcpConnection.disconnect();
+    public static Create(conn:TCP.Connection) { return new Session(conn); }
```
iislucas, line 257: +linebreak

src/socks-to-rtc/socks.ts
```diff
@@ -144,161 +146,167 @@ module Socks {
...
+      return conn.receive()
+          .then(Socks.Session.InterpretRequest)
+          .then(Socks.Session.CheckRequestFailure)
+          // Valid request - fire external callback.
+          .then((request) => {
+            return callback(this, request.addressString, request.port)
+          // Invalid request - notify client with |request.failure|.
+          }, (e) => {
+            replyToTCP(conn, parseInt(e.message));
+            return e;
```
iislucas: this should be a throw e.

src/socks-to-rtc/socks.ts
```diff
@@ -144,161 +146,167 @@ module Socks {
...
+          })
+          // Pass endpoint from external callback to client.
+          .then(Socks.Session.ComposeEndpointResponse)
+          .then((response) => { conn.sendRaw(response.buffer); })
+          .catch((e) => {
+            console.error(this + ': ' + e.message);
+            // TODO: Make disconnections server-handled to prevent
+            // bidirectional callbacks.
+            return e;
```
iislucas: throw again. :)

src/socks-to-rtc/socks.ts
```diff
@@ -144,161 +146,167 @@ module Socks {
...
     }
 
     // Given a data |buffer|, interpret the SOCKS request.
-    private handleRequest_ = (buffer) => {
-      // Only handle one request per tcp connection, so disable the |recv|
-      // callback for now. Pending data will be stored for the next handler.
-      this.tcpConnection.on('recv', null);
-
+    public static InterpretRequest = (buffer) => {
       var byteArray:Uint8Array = new Uint8Array(buffer);
```
iislucas: You don't need a type-annotation here. It is automatically types to Uint8Array by the new call.

src/socks-to-rtc/socks.ts
```diff
@@ -144,161 +146,167 @@ module Socks {
...
       }
+      return request;
+    }
 
-      // TODO: add a handler for failure to reach destination.
-      this.destinationCallback(
-          this,
-          this.request.addressString, this.request.port,
-          this.connectedToDestination_);
+    public static CheckRequestFailure(request) {
```
iislucas: does request have a type?

src/socks-to-rtc/tcp.ts
```diff
@@ -43,155 +50,182 @@ module TCP {
...
-                     this.serverSocketId + ' vs ' + acceptValue.serverSocketId);
-        return;
-      }
-      console.log('Tcp.Server accepted connection ' + acceptValue.clientSocketId);
-      var connectionsCount = Object.keys(this.openConnections).length;
-      if (connectionsCount >= this.maxConnections) {
-        fSockets.disconnect(acceptValue.clientSocketId);
-        fSockets.destroy(acceptValue.clientSocketId);
-        console.warn('TCP.Server: too many connections: ' + connectionsCount);
+    public isOn() { return this.serverSocketId > 0; }
```
iislucas: Can we change this to "isListening" or something? isOn is too ambiguous.

src/socks-to-rtc/tcp.ts
```diff
@@ -43,155 +50,182 @@ module TCP {
...
-                     this.connectionCallbacks);
+    /**
+     * Stop a TCP connection and remove from server.
+     */
+    public endConnection = (socketId) => {
+      console.log('ENDING ' + socketId);
+      this.conns[socketId]
+          .then(Connection.disconnect)
+          .then(this.removeFromServer_);
+      // TODO: Do we need an external callback here?
```
iislucas, line 226: Return the promise?

src/socks-to-rtc/tcp.ts
```diff
@@ -218,34 +252,36 @@ module TCP {
     // user might need (ie socketInfo). The socket shouldn't be doing work for
     // the user until the internals are ready.
     private initialized_:boolean = false;
+    public callbacks;
+
+    private disconnectPromise:Promise<void> = null;
+    private fulfillDisconnect = null;
+
+    // Static connection creation function which returns a promise.
+    static Create = (socketId, callbacks):Promise<Connection> => {
```
iislucas: Give a typing to the arguments, especially the callbacks. That's the easiest thing to get wrong.

src/socks-to-rtc/tcp.ts
```diff
@@ -279,16 +315,34 @@ module TCP {
     }
 
     /**
+     * Obtain a promise for a buffer as the result of a recv.
+     */
+    public receive = (minByteLength?:number):Promise<any> => {
```
iislucas: Promise should be given the correct type. A buffer presumably.

src/socks-to-rtc/tcp.ts
```diff
@@ -321,24 +375,20 @@ module TCP {
...
       fSockets.destroy(this.socketId);
-      // Make disconnect callback if not null
-      disconnectCallback && disconnectCallback(this);
-      // Fire removal callback for the Server containing this callback.
-      this.callbacks.removed(this);
+      this.fulfillDisconnect();  // Fire the disconnect Promise.
+      return this;
     }
+    public onDisconnect() { return this.disconnectPromise; }
+    public static disconnect(conn:Connection) { return conn.disconnect(); }
```
iislucas, line 394: Do we really need this? Seems like it's not doing anything.

src/socks-to-rtc/tcp.ts
```diff
@@ -321,24 +375,20 @@ module TCP {
...
       fSockets.disconnect(this.socketId);
       fSockets.destroy(this.socketId);
-      // Make disconnect callback if not null
-      disconnectCallback && disconnectCallback(this);
-      // Fire removal callback for the Server containing this callback.
-      this.callbacks.removed(this);
+      this.fulfillDisconnect();  // Fire the disconnect Promise.
+      return this;
     }
+    public onDisconnect() { return this.disconnectPromise; }
```
iislucas: Should be called "onceDisconnected"

src/socks-to-rtc/tcp.ts
```diff
@@ -434,4 +485,11 @@ module Util {
     f.readAsArrayBuffer(bb);
   }
 
+  /**
+   * Wrapper around creating a Promise rejection with Error.
+   */
+  export function Reject(msg:string) {
```
iislucas: Reject -> reject

iislucas:
Looking way way nicer! good work! We still have a bit to cleanup. Added comments, mostly style stuff.  Please fix and push to this pull-request.

src/interfaces/promise.d.ts
```diff
@@ -1,41 +1,59 @@
...
+declare class Promise<T> {
+
+  constructor (f:(fulfill:(t?:T)=>void,
+                  reject:(e:Error)=>void)=>void);
 
-  constructor (onFulfilled:(fulfillObj : F) => void,
-               onRejected:(rejectObj? : R) => void);
+  // |onTulfilled| either returns a promise...
+  then<T2> (fulfill:(t:T) => Promise<T2>,
+            reject?:(e:Error) => void)
```
iislucas: Just realised that the `reject` function should also return a "Promise<T2>"; it's supposed to recover from the rejection.

src/interfaces/promise.d.ts
```diff
@@ -1,41 +1,59 @@
...
-               onRejected?:(R) => void)
-      :Promise<F2,R2>;
+  // or the next fulfillment object directly.
+  then<T2> (fulfill?:(t:T) => T2,
+            reject?:(e:Error) => void)
+      :Promise<T2>;
 
-  catch (catchFn:(rejectObj:R) => void)
-      :Promise<F,R>;
+  catch (catchFn:(e:Error) => void)
```
iislucas: The catch function is supposed to return a new promise type too, so you can chain it with a then, so I think it should be `catchFn:(e:Error) => Promise<T>`

src/socks-to-rtc/socks-to-rtc.ts
keroserene: Indeed - it can never be null :) I'll update this

src/socks-to-rtc/socks.ts
keroserene: Going to return a Promise.reject here

src/socks-to-rtc/socks.ts
keroserene: Nothing at the moment

src/socks-to-rtc/tcp.ts
keroserene, line 394: It's just sugar which let's me do `.then(disconnect)` instead of `.then((conn:Connection) => { return conn.disconnect(); })` :)

src/socks-to-rtc/socks.ts
keroserene: not yet. There's a TODO above

src/socks-to-rtc/tcp.ts
keroserene: We could. Also it doesn't seem like this function is being called from anywhere so I'm actually going to get rid of it

src/interfaces/promise.d.ts
keroserene: It doesn't have to return a promise - it can also just return a value

PR #22, opened 2014-02-11
make sample curl command have the proxy resolve DNS

trevj:
Not completely sure why curl is having DNS resolution issues on my Linux box but this makes it work.


iislucas:
LGTM. Thanks for fix!

PR #24, opened 2014-02-12
promises - tcp and chrome sockets

keroserene:
- No longer have TcpEchoServer messages going to the wrong socket. (fix #5)
- Change to using typed promises for the read loop in Sockets.Chrome

Tested: Manual end-to-end.

### DIFF
--- a/src/chrome-fsocket.ts
+++ b/src/chrome-fsocket.ts
@@ -2,21 +2,9 @@
  * Chrome sockets over freedom sockets.
  * TODO: This should be refactored into freedom someday...
  */
+/// <reference path='interfaces/socket.d.ts' />
 
 declare var chrome:any;
-
-
-// Platform independent interface. TODO: Put somewhere so firefox can use it.
-interface ISockets {
-  create:any;
-  listen:any;
-  connect:any;
-  write:any;
-  getInfo:any;
-  disconnect:any;
-  destroy:any;
-  on?:any;
-}
 
 
 // http://developer.chrome.com/apps/socket.html
@@ -104,23 +92,24 @@ var ERROR_MAP = {
  * the socket user.
  */
 var readSocket = function(socketId) {
-  var saved_socketId = socketId;
-  var dataRead = (readInfo) => {
-    if (readInfo.resultCode > 0) {
-      var arg = {socketId: socketId, data: readInfo.data};
+  var dataRead = function (readInfo) {
+    var resultCode = readInfo.resultCode;
+    if (resultCode > 0) {
+      var arg = { socketId: socketId, data: readInfo.data };
       this.dispatchEvent('onData', arg);
       chrome.socket.read(socketId, null, dataRead);
-    } else if (0 === readInfo.resultCode) {
+      return;
+    }
+    var msg = '' + resultCode;
+    if (msg in ERROR_MAP) {
+        msg = ERROR_MAP[msg];
+    }
+    if (0 === resultCode) {
       console.log('socket ' + socketId + ' has been closed.')
-      this.dispatchEvent('onDisconnect', { socketId: socketId, error: msg });
     } else {
-      var msg = '' + readInfo.resultCode;
-      if (msg in ERROR_MAP) {
-          msg = ERROR_MAP[msg];
-      }
-      console.error('When reading from socket ' + socketId + ', return encounter error ' + msg);
-      this.dispatchEvent('onDisconnect', { socketId: socketId, error: msg });
-    };
-  };
+      console.error('socket ' + socketId + ': ' + msg);
+    }
+    this.dispatchEvent('onDisconnect', { socketId: socketId, error: msg });
+  }.bind(this);
   chrome.socket.read(socketId, null, dataRead);
 };
--- a/src/chrome-fsocket.ts
+++ b/src/chrome-fsocket.ts
@@ -3,9 +3,15 @@
  * TODO: This should be refactored into freedom someday...
  */
 /// <reference path='interfaces/socket.d.ts' />
+/// <reference path='interfaces/promise.d.ts' />
 
 declare var chrome:any;
 
+// https://developer.chrome.com/apps/socket.html#method-read
+interface ChromeReadInfo {
+  resultCode:number;
+  data:ArrayBuffer
+}
 
 // http://developer.chrome.com/apps/socket.html
 class ChromeSockets implements ISockets {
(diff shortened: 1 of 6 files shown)


src/chrome-fsocket.ts
```diff
@@ -2,125 +2,158 @@
...
-
+  // https://developer.chrome.com/apps/socket.html#method-read
+  interface ChromeReadInfo {
+    resultCode:number;
+    data:ArrayBuffer
+  }
 
-// http://developer.chrome.com/apps/socket.html
-class ChromeSockets implements ISockets {
+  // http://developer.chrome.com/apps/socket.html
```
iislucas: The URL comment here is a little odd, can we clarify this is where to find the chrome socket spec, but this is not the interface that this class provides.

src/chrome-fsocket.ts
```diff
@@ -2,125 +2,158 @@
...
+    }
 
-  public listen = (socketId, address, port, callback) => {
-    chrome.socket.listen(socketId, address, port, null, (result) => {
-      callback(result);
-      if (0 === result) {
-        var acceptCallback = function (acceptInfo) {
+    public listen = (socketId, address, port, callback) => {
+      chrome.socket.listen(socketId, address, port, null, (result) => {
+        callback(result);
```
iislucas, line 41: Given that an error can happen, I'm not sure we want to do the callback here. Isn't it better to do the callback at the end, when you know that listen has actually completed...? What do you think?

src/chrome-fsocket.ts
```diff
@@ -2,125 +2,158 @@
...
+        this.doReadLoop_(socketId);
+      });
+    }
 
-  public listen = (socketId, address, port, callback) => {
-    chrome.socket.listen(socketId, address, port, null, (result) => {
-      callback(result);
-      if (0 === result) {
-        var acceptCallback = function (acceptInfo) {
+    public listen = (socketId, address, port, callback) => {
```
iislucas, line 39: we sometimes call the argument "callback" and sometimes "continuation". Lets check the freedom convention, and use that consistently? :)

src/chrome-fsocket.ts
```diff
@@ -2,125 +2,158 @@
...
-    });
-  }
+      continuation();
+    }
 
-  public destroy = (socketId, continuation) => {
-    if (chrome && chrome.socket) {
-      chrome.socket.destroy(socketId);
+    public disconnect = (socketId:number, continuation) => {
+      if (chrome && chrome.socket) {
```
iislucas: What's with the chrome & chrome.socket  check? None of this code will work without it defined. Maybe remove?

src/chrome-fsocket.ts
```diff
@@ -2,125 +2,158 @@
...
+     * the socket user.
+     */
+    private doReadLoop_ = (socketId:number) => {
+      return this.promiseRead_(socketId)
+          .then(this.checkResultCode_)
+          .then((data) => {
+            // This still dispatches to *all* handlers attached to onData, and
+            // puts the responsibility on the user of this object to act only for
+            // the socket corresponding to |socketId|. Really bad.
+            // TODO: Make the events a bijection.
```
iislucas: I think what we should do here is have sockets wrapper code for module using sockets. That can manage the bijection. That means we write the code just once, and we still get simple interfaces. We then only have a single thing bound on the other side (as in, in the module using sockets)

src/chrome-fsocket.ts
```diff
@@ -2,125 +2,158 @@
...
+            this.fireEvent('onData', {
+              socketId: socketId,
+              data: data
+            });
+            return socketId;
+          })
+          .then(this.doReadLoop_)
```
iislucas: I don't think this is good... each loop of this will add a catch handler to the stack, and it will then crash once the stack gets too big. Better to use a while loop, or explicit call at the end, and have the catch before this line.

src/chrome-fsocket.ts
```diff
@@ -2,125 +2,158 @@
...
-    } else if (0 === readInfo.resultCode) {
-      console.log('socket ' + socketId + ' has been closed.')
-      this.dispatchEvent('onDisconnect', { socketId: socketId, error: msg });
-    } else {
-      var msg = '' + readInfo.resultCode;
-      if (msg in ERROR_MAP) {
+    /**
+     * Create a promise for a future reading of this socket.
+     */
+    private promiseRead_ = (socketId:number):Promise<ChromeReadInfo> => {
```
iislucas: cute.

src/chrome-fsocket.ts
```diff
@@ -2,125 +2,158 @@
...
+     * Othrerwise, pass along read data.
+     */
+    private checkResultCode_ = (readInfo:ChromeReadInfo) => {
+      var code = readInfo.resultCode;
+      if (code <= 0) {
+        var msg = '' + code;
+        if (msg in ERROR_MAP) {
           msg = ERROR_MAP[msg];
+        }
+        if (0 === code) {
```
iislucas: Move this if above the if(code <= 0), and make it an if(code < 0) ...
One fewer conditional checks, and less lookup for the most common case. Also avoids nested if's when they can easily be flat.

src/socks-to-rtc/tcp.ts
```diff
@@ -50,8 +47,12 @@ module TCP {
 
     // TODO: replace with promises.
     private callbacks:any;
-    private connectionCallbacks:IConnectionCallbacks;
 
+    /**
+     * Create TCP server.
+     * @param {Object} options Options of the form { maxConnections: integer,
```
iislucas: I'd rather have a typescript interface if we know the form that the function assumes the input will take. :) 
For typescript code, we should provide types, not closure style comments.

src/socks-to-rtc/tcp.ts
```diff
@@ -50,8 +47,12 @@ module TCP {
...
+     * allowHalfOpen: bool }.
+     */
     constructor(public addr, public port, options?) {
```
iislucas: you know how I love types...

src/socks-to-rtc/tcp.ts
keroserene: This comment is a relic from somewhere else :) but I can make it a type

src/socks-to-rtc/tcp.ts
```diff
@@ -356,7 +347,7 @@ module TCP {
      * @param {String} msg The message to send.
      * @param {Function} callback The function to call when the message has sent.
```
iislucas: For next patch: move types into typescript, out of comments.

src/socks-to-rtc/tcp.ts
```diff
@@ -356,7 +347,7 @@ module TCP {
...
      */
-    public send(msg:string, callback?) {
+    public send = (msg:string, callback?) => {
```
iislucas: Why this style?

iislucas:
Added a bunch of comments; can be for next patch. LGTM.

src/socks-to-rtc/tcp.ts
keroserene: Makes sure 'this' is bound to the class. The other syntax doesn't enforce it, especially if the method is passed as a callback

PR #26, opened 2014-02-13
Freedom upgrade

trevj:
Had a hunch this wasn't necessary...all seems to work fine without.

PR #28, opened 2014-02-13
remove unnecessary include

trevj:
Oops -- please ignore my earlier pull request (https://github.com/uProxy/socks-rtc/pull/26).

This removes what I think is an unnecessary include because the socks-to-rtc Freedom module also pulls in tcp.js. Tested locally with curl running against the Chrome app.

### DIFF
--- a/chrome/socks_rtc.json
+++ b/chrome/socks_rtc.json
@@ -3,7 +3,6 @@
   "description": "socks-rtc Freedom manifest",
   "app": {
     "script": [
-      "js/socks-to-rtc/tcp.js",
       "smell_socks.js"
     ]
   },


keroserene:
The tcp.ts file is there in order to test the TcpEchoServer from the chrome app now and then, because we still haven't gotten freedom to work with jasmine. This line can be removed once we get the test rig actually set up.

Also, can we hold off on miniature pull requests like this, because they inflate the git commit history with many "merge" messages? There are other ongoing branches, and git has the tendency to get messy, fast ;)

iislucas:
I think small CLs are fine. We should just make them branches off of the
origin master, and do the appropriate rebasing. I'll try and write more in
the coding guide for it.

This patch is good, but we probably want to merge after we have finished
whatever testing we're currently doing. Or we want a separate json/html for
the tcp server test, so we don't mis them up. Generally having a simple TCP
server echo-test is a good thing too.

I'll create a new issue... https://github.com/uProxy/socks-rtc/issues/29

On Thu, Feb 13, 2014 at 4:06 PM, Serene H notifications@github.com wrote:

> The tcp.ts file is there in order to test the TcpEchoServer from the
> chrome app now and then, because we still haven't gotten freedom to work
> with jasmine. This line can be removed once we get the test rig actually
> set up.
> 
> Also, can we hold off on miniature pull requests like this, because they
> inflate the git commit history with many "merge" messages? There are other
> ongoing branches, and git has the tendency to get messy, fast ;)
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uProxy/socks-rtc/pull/28#issuecomment-35025805
> .

## 

Lucas Dixon | Google Ideas

trevj:
I don't feel strongly about this change at all so that's fine but I think short changes are often preferable (to be more specific: as self-contained possible, which is often the same thing) and, in any case, unavoidable unless we expect to get everything right first time (which I know I can't) :-)

I'll read up on rebasing!

PR #31, opened 2014-02-14
travis config with a jasmine test for socks.ts

trevj:
This adds a config for Travis and a near-trivial test for socks.ts, which I picked because it has a few functions without any dependencies (no need for the DOM or to load Freedom, etc.). It's just a start to get us setup with a continuous build!

Caveats:
- no Selenium tests yet (that's a whole different kettle of balls of wax and I don't want it to block us adding more unit tests)
- this doesn't test any Freedom-dependant code...hopefully that won't be hard
- we'll probably want to write tests in TypeScript at some point...I don't think anything here precludes that

There are several commits in this branch so I'll be sure to rebase rather than merging to keep the history clean :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,5 @@ language: node_js
 node_js:
  - "0.11"
 before_install:
- - npm install -g grunt grunt-cli
+ - npm install -g grunt-cli
  - grunt setup
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: node_js
 node_js:
  - "0.11"
+ - "0.10"
+ - "0.8"
+ - "0.6"
 before_script:
  - npm install -g grunt-cli
  - grunt setup
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,5 @@ language: node_js
 node_js:
  - "0.11"
 before_install:
- - npm install -g grunt-cli
+ - npm install -g grunt grunt-cli
  - grunt setup
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
  - "0.11"
-before_install:
+before_script:
  - npm install -g grunt-cli
  - grunt setup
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -34,4 +34,4 @@ For example:
 
 `curl -x socks5h://localhost:9999 www.google.com`
 
-There will be actual tests soon!
+There will be more tests soon!
```
keroserene, line 37: yay!

spec/socks-to-rtc/socks_spec.js
```diff
@@ -0,0 +1,43 @@
...
+    ipv4Request = new Uint8Array([
+      Socks.VERSION5,
+      Socks.REQUEST_CMD.CONNECT,
+      0, // reserved
+      Socks.ATYP.IP_V4,
+      192, 168, 1, 1, // IP: 192.168.1.1
+      1200 >> 8, 1200 & 0xFF]); // port: 1200
+  });
+
+  it("reject wrongly sized requests", function() {
```
keroserene: nit: We should prefer single quotes over double quotes in most cases

keroserene:
Really cool - just that one nit, and LGTM / feel free to merge!

.travis.yml
```diff
@@ -0,0 +1,6 @@
+language: node_js
+node_js:
+ - "0.11"
```
willscott, line 3: Probably worth testing the current stable "0.10"  as well. maybe try previous stable "0.8" version as well.  Having multiple runs per commit helps detect nondeterminism / flakyness.

src/socks-to-rtc/socks.ts
```diff
@@ -88,10 +90,14 @@ module Socks {
    * Parse byte array into a SOCKS request object.
    */
   export function interpretSocksRequest(byteArray:Uint8Array) {
+    var result:any = {};
```
iislucas, line 93: add TODO: for specifying the result's real type.

.travis.yml
trevj, line 3: Done!

PR #32, opened 2014-02-14
update to freedom 2.1

keroserene:
Finally works! Many things had to change, like explicitly creating data channels.
- better peerconnection / signalling channel interactions
- promises for the data channels
- more communications interface types
- after extensive debugging, maybe we should try the freedom transport interface - see (https://github.com/UWNetworksLab/freedom/issues/42)

Tested: end-to-end

### DIFF
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -35,7 +35,8 @@ module RtcToNet {
       // Create a signalling channel.
       fCore.createChannel().done((chan) => {
         // this.sctpPc.setup(chan.identifier, 'RtcToNet-' + this.peerId, false);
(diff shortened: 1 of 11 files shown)


keroserene:
Note: monotonically-increasing duplicate datachannel messages from issue #30 still seems to be here. Need to determine if there's some other silly culprit, or if it's actually just something which forgoes clean solutions

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -77,42 +60,61 @@ module RtcToNet {
       for (var i in this.netClients) {
         this.netClients[i].close();  // Will close its data channel.
       }
-      this.sctpPc.close();
+        this.sctpPc.close();
```
iislucas, line 63: odd space?

iislucas:
Nice work!

PR #33, opened 2014-02-14
disconnections - sockts-to-rtc pathway

keroserene:
- Fixed missing 'onDisconnect' handler in `core.socket`
- There will be another pull request double checking the rtc-to-net pathways.
- onCloseDataChannel appears to not be firing on the remote peer as well...

Tested: end-to-end


keroserene:
Removed the `freedom_bugfix.js` because it's updated to freedom 2.2

iislucas:
LGTM

PR #36, opened 2014-02-15
And an end-to-end Selenium-based test

trevj:
This adds a test which:
- starts Chrome, with the app installed and the browser's proxy settings pointed to localhost:9999
- fetches a few pages and verifies they load

I added m.guardian.co.uk because right now it _doesn't_ load via the proxy. No idea why!

Caveats:
- requires the selenium server to be running locally (there are a few grunt plugins which purport to do this but I couldn't get them to work)
- no Travis/Sauce Labs connection yet...uploading the extension will take a little more work (it seems to be possible but let's take this in stages, there's other things to do too!)

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -1,5 +1,7 @@
 module.exports = (grunt) ->
 
+  path = require('path');
+
   grunt.initConfig {
     pkg: grunt.file.readJSON('package.json'),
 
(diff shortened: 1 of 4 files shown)


iislucas:
nice! Looks like you need to pull/rebase and then feel free to merge after a manual test.

trevj:
Apologies for the delay in merging; should be good to pull now.

keroserene:
Looks great!

PR #37, opened 2014-02-16
disconnections - end-to-end works both ways

keroserene:
Use `msg.text` as payload to signal disconnections across the datachannel.
- `SOCKS-DISCONNECT` occurs when the SocksToRtc's TCP request socket disconnects, ending the SOCKS session. This signal crosses the data channel to close the RtcToNet's client socket.
- `NET-DISCONNECT` occurs when the RtcToNet socket connection to destination closes first. This signal crosses the data channel to close the SocksToRtc side SOCKS session / TCP connection.
- Fixed _local_ vs. _remote_ disconnection semantics. The `onceDisconnected`-like promises only fulfill for remote disconnection events. This prevents cyclical/redundant calls to disconnection handlers & checks, since the closing code paths can occur in either direction.

Eventually, when the freedom datachannels properly send closes across to the remote peer, this may be removed.

Tested: End-to-end

### DIFF
--- a/src/rtc-to-net/netclient.ts
+++ b/src/rtc-to-net/netclient.ts
@@ -33,8 +33,8 @@ module Net {
     private queue:any[] = [];
     private state:State = State.CLOSED;
 
-    private closePromise:Promise<void>;
-    private fulfillClose:()=>void;
+    private disconnectPromise:Promise<void>;
+    private fulfillDisconnect:()=>void;
 
     /**
      * Constructing a Net.Client immediately begins a socket connection.
@@ -44,8 +44,8 @@ module Net {
         private onResponse:(buffer:ArrayBuffer)=>any,
         private destination:Destination) {
       this.state = State.CREATING_SOCKET;
-      this.closePromise = new Promise<void>((F, R) => {
-        this.fulfillClose = F;  // To be fired on close.
+      this.disconnectPromise = new Promise<void>((F, R) => {
+        this.fulfillDisconnect = F;  // To be fired on close.
       });
       this.createSocket_()  // Initialize client TCP socket.
           .then(this.connect_)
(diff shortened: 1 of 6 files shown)


src/rtc-to-net/netclient.ts
```diff
@@ -151,26 +151,34 @@ module Net {
     }
 
     /**
-     * Close socket and fire external close handlers.
+     * Fired by closes remotely.
```
iislucas: fix comment. :)

src/rtc-to-net/netclient.ts
```diff
@@ -69,7 +69,13 @@ module Net {
       }
     }
 
-    public close = () => { this.onClose_ };
+    /**
+     * Close the Net.Client locally.
+     */
+    public close = () => {
```
iislucas, line 75: Not clear to me that you need a closeSocket_ and a close. Why not just put code from closeSocket in here, and make all calls go to close?

iislucas:
Noticed that attach_handlers_ has the error checking code in it. I think it would be better for that code to be in a separate then, or the error is raised as a promise failure earlier. It's not right for attach_handlers_ to do the error raising/promise failing. 

(sorry I should have spotted this earlier)

src/socks-to-rtc/tcp.ts
```diff
@@ -384,11 +385,16 @@ module TCP {
       // Close the socket.
       fSockets.disconnect(this.socketId);
       fSockets.destroy(this.socketId);
-      this.fulfillDisconnect(0);  // Fire the disconnect Promise.
       return this;
     }
     public static disconnect(conn:Connection) { return conn.disconnect(); }
 
+    // Fire the disconnect Promise.
+    public fireRemoteDisconnect = () => { this.fulfillDisconnect(0); }
```
iislucas: why public?

iislucas:
Generally looks good. a couple of small things. When fixed and tested again, merge away!

src/socks-to-rtc/tcp.ts
keroserene: Since the freedom handler for disconnections fires generically with a socketInfo argument, TCP.Server needs to determine which socket disconnected.

To keep promises chained in the same place, and also maintain a distinction between local closes and remote disconnections without callback soup, it seems the best approach was to fulfill the connection disconnect promise from outside :/

However, I have still revised the chaining here to actually reflect close vs disconnect

PR #40, opened 2014-02-21
Switch to freedom-runtime-chrome dependency

willscott:
Hopefully simplifies things.
Also will allow removal of chrome/lib/boilerplate.js and src/chrome-fsocket.ts

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,3 @@ node_js:
  - "0.8"
 before_script:
  - npm install -g grunt-cli
- - grunt setup
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -14,7 +14,7 @@ module.exports = (grunt) ->
             expand: true,
             cwd: 'build'
           }, {
-            src: 'node_modules/freedom/freedom.js'
+            src: 'node_modules/freedom-runtime-chrome/freedom.js'
             dest: 'chrome/js/freedom.js'
           }, {
             src: ['**/*.json']
@@ -60,17 +60,6 @@ module.exports = (grunt) ->
       }
     }
 
-    shell: {
-      freedom_setup: {
-        command: 'npm install',
-        options: {stdout: true, stderr: true, failOnError: true, execOptions: {cwd: 'node_modules/freedom'}}
-      }
-      freedom_build: {
-        command: 'grunt',
-        options: {stdout: true, stderr: true, failOnError: true, execOptions: {cwd: 'node_modules/freedom'}}
-      }
-    }
-
     jasmine: {
       # Eventually, this should be a wildcard once we've figured out how to run
       # more dependencies under Jasmine.
@@ -102,7 +91,6 @@ module.exports = (grunt) ->
   grunt.loadNpmTasks 'grunt-contrib-copy'
   grunt.loadNpmTasks 'grunt-contrib-clean'
   grunt.loadNpmTasks 'grunt-contrib-jasmine'
-  grunt.loadNpmTasks 'grunt-shell'
   grunt.loadNpmTasks 'grunt-ts'
   grunt.loadNpmTasks 'grunt-jasmine-node'
   grunt.loadNpmTasks 'grunt-env'
@@ -137,11 +125,4 @@ module.exports = (grunt) ->
   grunt.registerTask 'chrome', [
     'build',
     'copy:app'
-  ]
-
-  # Freedom doesn't build correctly by itself - run this task when in a clean
-  # directory.
-  grunt.registerTask 'setup', [
-    'shell:freedom_setup',
-    'shell:freedom_build'
   ]
--- a/README.md
+++ b/README.md
@@ -22,13 +22,12 @@ webserver, and serves the response back to _socks-to-rtc_.
 #### Requirements
 
 - node + npm
-- grunt `npm install -g grunt-cli`
+- `npm install -g grunt-cli`
+- `grunt`
 
 #### Initial Setup
 
 - Run `npm install` from the base directory to obtain all prerequisites.
-- Run `grunt setup` to prepare the freedom installation. (This step should
-  disappear with future versions of freedom)
 
 #### Build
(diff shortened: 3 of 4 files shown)


iislucas:
Nice! did you check that end-to-end test still works?

willscott:
Scroll down on the page :)

https://github.com/uProxy/socks-rtc/pull/40

keroserene:
I believe the travis tests don't include automated end-to-end via selenium yet. You still need to do it manually :)

willscott:
You're correct.  This isn't working :)

The bug when I try to run it is that this dep upgrades us to freedom v0.3 which returns promises.  We'll get to strip out the typescript wrappers around the 'done()'s in the modules as part of this!

keroserene:
Ooh, excellent. I can make those commits to this pull request, if you'd like.

Also, I'm in the middle of plugging socks-rtc into uproxy. If we upgrade this to freedom 0.3, does this imply upgrading uproxy to 0.3 as well? Will everything (like the social/xmpp stuff) work?

willscott:
Sounds good!

Also yes on the 0.3  -> it just changes '.done' to '.then'

--Will

iislucas:
I like the idea to adapt the current Freedom2 branch of uProxy to Freedom3,
and update it as needed for changes in the social API.

+uproxy-eng
Do we have a short outline of what changed in the social API? Or even just
how the new one should be used and what values we should expect? (I'm happy
to turn such a thing into a documented typescript interface, which provides
a more formal and accurate technical description)

On Sat, Feb 22, 2014 at 3:21 PM, Will notifications@github.com wrote:

> Sounds good!
> 
> Also yes on the 0.3 -> it just changes '.done' to '.then'
> 
> --Will
> 
> On Sat, Feb 22, 2014 at 11:41 AM, Serene H notifications@github.com
> wrote:
> 
> > Ooh, excellent. I can make those commits to this pull request, if you'd
> > like.
> > 
> > Also, I'm in the middle of plugging socks-rtc into uproxy. If we upgrade
> > this to freedom 0.3, does this imply upgrading uproxy to 0.3 as well?
> > Will
> > everything (like the social/xmpp stuff) work?
> > 
> > ## 
> > 
> > Reply to this email directly or view it on GitHub<
> > https://github.com/uProxy/socks-rtc/pull/40#issuecomment-35812224>
> > 
> > .
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uProxy/socks-rtc/pull/40#issuecomment-35813501
> .

## 

Lucas Dixon | Google Ideas

keroserene:
`Socket_chrome` included in `freedom_runtime_chrome` is really old compared to `chrome_fsockets.ts`. I'd rather not rollback and lose the typing, error handling, and easier-to-read internal promise logic... can we typescript the chrome runtime?

willscott:
Had this conversation with Lucas over the weekend. I'm don't want the
typescript in the non-uproxy-specific freedom code. Lets take a look and
figure out what got done in the ts version that isn't in the javascript one.

--Will

keroserene:
It maps error codes to messages, provides return types (like `Socket.CreateInfo, .DisconnectInfo, .ReadInfo`) and has clearer readloop logic using promises. Of course it's not actually a big deal. I just find it silly to rollback and lose this work.

willscott:
Let me know if any changes (other than types :)) aren't now reflected.

https://github.com/freedomjs/freedom-runtime-chrome/blob/master/providers/socket.js

--Will

iislucas:
Why not just use the JS compiled version (with comments) of
chrome_fsockets? That will have all the nice new work and it will match the
correct typing?

On Mon, Feb 24, 2014 at 6:09 PM, Will notifications@github.com wrote:

> Let me know if any changes (other than types :)) aren't now reflected.
> 
> https://github.com/freedomjs/freedom-runtime-chrome/blob/master/providers/socket.js
> 
> --Will
> 
> On Mon, Feb 24, 2014 at 2:02 PM, Serene H notifications@github.com
> wrote:
> 
> > It maps error codes to messages, provides return types (like
> > Socket.CreateInfo,
> > .DisconnectInfo, .ReadInfo) and has clearer readloop logic using
> > promises. Of course it's not actually a big deal. I just find it silly to
> > rollback and lose this work.
> > 
> > ## 
> > 
> > Reply to this email directly or view it on GitHub<
> > https://github.com/uProxy/socks-rtc/pull/40#issuecomment-35944920>
> > 
> > .
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uProxy/socks-rtc/pull/40#issuecomment-35951962
> .

## 

Lucas Dixon | Google Ideas

willscott:
No direct reason, i guess?
1. this took 20 minutes
2. i could write the comments in the yuidocs style that the rest of freedom
   is in
3. I could transform the ERROR_MAP into a property on the object, rather
   than as a separate global var.
4. I could keep top jslint comments that make it validate in the stricter
   jslint checker.

--Will

iislucas:
On Mon, Feb 24, 2014 at 7:02 PM, Will notifications@github.com wrote:

> No direct reason, i guess?
> 1. this took 20 minutes
> 2. i could write the comments in the yuidocs style that the rest of freedom
>    is in
> 3. I could transform the ERROR_MAP into a property on the object, rather
>    than as a separate global var.
> 4. I could keep top jslint comments that make it validate in the stricter
>    jslint checker.

fair enough. I guess we can make the typescript interface for this as a
separate file.

## 

Lucas Dixon | Google Ideas

willscott:
Typescript interface should be the freedom.socket api, rather than the
class itself, right? That's the point of the indirection through freedom :)

--Will

iislucas:
On Mon, Feb 24, 2014 at 7:22 PM, Will notifications@github.com wrote:

> Typescript interface should be the freedom.socket api, rather than the
> class itself, right? That's the point of the indirection through freedom :)

We lost of the stronger guarantee of correctness in chrome_sockets by
dropping TS from it, so of course ideally you'd have both. But you're quite
right that what we really want is the TS interface for the
freedom-indirection interface of sockets api :)

PR #41, opened 2014-02-26
Yangoon core.udpsocket

trevj:
This adds a provider for core.udpsocket:
UWNetworksLab/freedom#52

Based quite heavily on chrome-fsocket.ts.

Also some bits and bobs of plumbing which aren't strictly necessary right now but will be useful for the next merge, which will consist of a "UDP server" (basically, a counterpart of tcp.ts).

This works pretty well with some toy code I have which adds UDP support to the proxy (pretty far from ready for review).

There's also a test. It's far from bulletproof but it does check that we're making the right calls to chrome.sockets.*.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -117,7 +117,7 @@ module.exports = (grunt) ->
   # This is the target run by Travis. Targets in here should run locally
   # and on Travis/Sauce Labs.
   grunt.registerTask 'test', [
-    'build',
+    'chrome',
     'jasmine'
   ]
 
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -74,7 +74,10 @@ module.exports = (grunt) ->
     jasmine: {
       # Eventually, this should be a wildcard once we've figured out how to run
       # more dependencies under Jasmine.
-      src: 'chrome/js/socks-to-rtc/socks.js',
+      src: [
+        'chrome/js/socks-to-rtc/socks.js',
+        'chrome/js/chrome-udpsocket.js'
+      ],
       options : {
         specs : 'spec/**/*_spec.js'
       }
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -53,8 +53,8 @@ module.exports = (grunt) ->
           sourceMap: false
         }
       }
-      chromeFSocket: {
-        src: ['src/chrome-fsocket.ts'],
+      chromeSockets: {
+        src: ['src/chrome-*socket.ts'],
         outDir: 'build/',
         options: { sourceMap: false; }
       }
@@ -110,7 +110,7 @@ module.exports = (grunt) ->
   grunt.registerTask 'build', [
     'ts:socks2rtc',
     'ts:rtc2net',
-    'ts:chromeFSocket',
+    'ts:chromeSockets',
     'copy:json'
   ]
 
--- a/chrome/lib/boilerplate.js
+++ b/chrome/lib/boilerplate.js
@@ -4,4 +4,5 @@
 window.freedomcfg = function(register) {
   // Necessary so we can actually use chrome sockets.
   register('core.socket', Sockets.Chrome);  // src/chrome-fsocket.ts
+  register('core.udpsocket', UdpSocket.Chrome);  // src/chrome-udpsocket.ts
 }
--- a/chrome/socks_rtc.html
+++ b/chrome/socks_rtc.html
@@ -3,6 +3,7 @@
   <head>
     <title>socks-rtc</title>
     <script type='text/javascript' src='js/chrome-fsocket.js'></script>
+    <script type='text/javascript' src='js/chrome-udpsocket.js'></script>
     <script type='text/javascript' src='lib/boilerplate.js'></script>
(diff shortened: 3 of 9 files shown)


chrome/lib/boilerplate.js
```diff
@@ -4,4 +4,5 @@
 window.freedomcfg = function(register) {
   // Necessary so we can actually use chrome sockets.
   register('core.socket', Sockets.Chrome);  // src/chrome-fsocket.ts
```
iislucas, line 6: In a followup CL, make this "tcpsocket".

Gruntfile.coffee
```diff
@@ -110,14 +113,14 @@ module.exports = (grunt) ->
   grunt.registerTask 'build', [
     'ts:socks2rtc',
     'ts:rtc2net',
-    'ts:chromeFSocket',
+    'ts:chromeSockets',
     'copy:json'
   ]
 
   # This is the target run by Travis. Targets in here should run locally
```
iislucas, line 120: I think might be better to organize tasks so that the sub-tasks a task depends on are listed before the task itself, so that you know which way in the file to start you search :)

chrome/socks_rtc.html
```diff
@@ -3,6 +3,7 @@
   <head>
     <title>socks-rtc</title>
     <script type='text/javascript' src='js/chrome-fsocket.js'></script>
```
iislucas, line 5: in follow-up CL rename 'chrome-fsocket.js' to 'chrome-tcpsocket.js'

src/chrome-udpsocket.ts
```diff
@@ -0,0 +1,82 @@
...
+  interface CreateSocketInfo {
+    socketId:number;
+  }
+
+  export class Chrome implements UdpSocket.API {
+    private socketId:number;
+
+    constructor (
+        private channel,
+        private dispatchEvent:(event:string, data:any) => void) {
```
iislucas, line 37: Maybe better to provide a more specific typing for data. You have an interface that provides a specific typing for that data, so best to make an interface for it, I think you already defined "RecvFromInfo".  I think having some more description of the kinds of events is a good idea. What do you think?

src/chrome-udpsocket.ts
```diff
@@ -0,0 +1,82 @@
...
+        this.recvFromLoop();
+      });
+    }
+
+    public sendTo = (data:ArrayBuffer, address:string, port:number, continuation) => {
+      // TODO(yangoon): throw error if socketId unset.
+      chrome.socket.sendTo(this.socketId, data, address, port, (writeInfo) => {
+        console.log(writeInfo);
+      });
+      continuation();
```
iislucas: I think the continuation should be called when data is sent, i.e. in the callback.

src/chrome-udpsocket.ts
```diff
@@ -0,0 +1,82 @@
+/**
+ * Freedom UDP sockets over the Chrome APIs.
+ */
+/// <reference path='interfaces/udpsocket.d.ts' />
+
+declare var chrome:any;
```
iislucas, line 7: I think we only use chrome.socket, so maybe declare that, and make a TODO to use the DefinitelyTyped interface for chrome.socket?

src/interfaces/udpsocket.d.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  export interface RecvFromInfo {
+    resultCode:number;
+    address:string;
+    port:number;
+    data:ArrayBuffer
+  }
+
+  export interface API {
+    bind:any;
```
iislucas, line 8: Provide real types :)

iislucas:
cool! a few comments that should be easy to fix.

chrome/lib/boilerplate.js
trevj, line 6: Sure. I'll follow up separately asking for the safest way to do this.

chrome/socks_rtc.html
trevj, line 5: Will do.

Gruntfile.coffee
trevj, line 120: Totally agree. Perhaps something for a followup CL?

iislucas, line 120: sure.

src/chrome-udpsocket.ts
trevj: Oops, big mistake. Thanks. Fixed now, and I updated the test to make sure that the number of bytes written is returned to the client.

src/chrome-udpsocket.ts
trevj, line 7: I wasn't able to figure out how to do something like:
declare var chrome.socket:any

so I've added a TODO (hopefully will follow up on this soon, so it won't be an issue for long).

src/chrome-udpsocket.ts
trevj, line 37: Done. I added types for the chrome.socket.\* callbacks and used them where appropriate and also added types for the continuations.

src/interfaces/udpsocket.d.ts
trevj, line 8: Done.

PR #43, opened 2014-03-05
accept UDP_ASSOCIATE requests and add a skeleton relay server

trevj:
This adds a class which will form the basis of our UDP support, plus some associated (and occasionally weird -- there are many TODOs in this) plumbing.

Some refactoring required before this is fully operational but with this our proxy server will accept UDP clients and provide them with a place to send their datagrams.

This requires core.udpsocket so it requires Freedom 0.3 -- but I'd appreciate any feedback you have now!


src/socks-to-rtc/udprelay.ts
```diff
@@ -0,0 +1,131 @@
...
+  *     client)
+   *  - this class will forward a copy of each datagram received on the socket
+   *    to the relevant remote host (minus the SOCKS5 headers, of course)
+   *  - this class will relay each datagram reply received from the remote
+   *    server *back* to the SOCKS5 client (prepended with the SOCKS5 headers,
+   *    of course)
+   *  - call destroy to clean up, typically when the TCP connection on which
+   *    the UDP_ASSOCIATE was negotiated is terminated (this is important
+   *    because relays are relatively expensive in terms of the number of
+   *    sockets required)
```
iislucas: Be precise: twice as many sockets. :)

src/socks-to-rtc/udprelay.ts
```diff
@@ -0,0 +1,131 @@
...
+   * A UDP-based "relay" server intended for use as part of a SOCKS5 proxy.
+   *
+   * Briefly, here's how to use this:
+   *  - create an instance of this class
+   *  - call bind, probably specifying port zero to have the system pick a
+   *    free port
+   *  - call getInfo, to discover on which port the relay is listening
+   *  - (the caller can now return the address and port back to the SOCKS
+  *     client)
+   *  - this class will forward a copy of each datagram received on the socket
```
iislucas, line 17: Maybe worth mentioning that the datagrams received are assumed to be SOCKS5 UDP requests?

src/socks-to-rtc/udprelay.ts
```diff
@@ -0,0 +1,131 @@
...
+   * 
+   * Other notes:
+   *  - while the RFC states that the relay MUST drop any message originating
+   *    from an IP other than that which requested the association, this
+   *    implementation makes no effort to do so (this isn't urgent because we
+   *    typically only listen locally)
+   *  - similarly, we completely ignore the BND.PORT and BND.ADDR fields
+   *    specified by the client, if any, during the handshake (it seems that
+   *    clients typically don't specify this anyway and, again, we're typically
+   *    only listening locally)
```
iislucas: add as TODO. We will need this later, I think, in order to be able to de-obfuscate. Chrome WebRTC will need to ask a SOCKS proxy to listen on it's behalf in order to handle incoming UDP streams. Those incoming streams currently go directly to chrome, but we will want to be able to handle them so that we can de-obfuscate before giving them to chrome.

src/socks-to-rtc/udprelay.ts
```diff
@@ -0,0 +1,131 @@
+module Socks {
+  /**
+   * A UDP-based "relay" server intended for use as part of a SOCKS5 proxy.
```
iislucas: add a link to the RFC?

src/socks-to-rtc/udprelay.ts
```diff
@@ -0,0 +1,131 @@
...
+   *
+   * Briefly, here's how to use this:
+   *  - create an instance of this class
+   *  - call bind, probably specifying port zero to have the system pick a
```
iislucas: Why 'probably' here and not below. Maybe remove 'probably' here?

src/socks-to-rtc/udprelay.ts
```diff
@@ -0,0 +1,131 @@
...
+    constructor () {
+      this.socket = freedom['core.udpsocket']();
+    }
+
+    /**
+     * Returns a promise to create a socket, bind to the specified address and
+     * port, and start relaying events. Specify port zero to have the system
+     * choose a free port.
+     */
+    public bind(
```
iislucas: I've added a section in our style guide on spaces and type annotations. 
https://docs.google.com/a/google.com/document/d/1ZV70oTNWO00bs0RcMwqr3FvKJXA6smYQ2kZ8Zr8YZnw/edit#heading=h.gmob2htqhzm4

I think this can all fit on one line.

src/socks-to-rtc/udprelay.ts
```diff
@@ -0,0 +1,131 @@
...
+        address:string,
+        port:number):Promise<any> {
```
iislucas: When there is no return type, use "void" instead of any. That way Typescript will give you an error if you try to bind a variable to the return type. i.e. here return type should be `Promise<void>`

src/socks-to-rtc/udprelay.ts
trevj: I clarified this (it's actually one to communicate with the client and one for each remote host with which it wishes to communicate (we can probably be smarter about that in time but I'm just aiming to make this all work for now))

src/socks-to-rtc/udprelay.ts
trevj, line 17: Yes!

src/socks-to-rtc/udprelay.ts
trevj: Sorry, GitHub isn't clear on exactly which line you meant -- add a TODO to do what?

src/socks-to-rtc/udprelay.ts
trevj: Done.

src/socks-to-rtc/udprelay.ts
trevj: Clarified.

src/socks-to-rtc/udprelay.ts
trevj: Good to know, thanks!

src/socks-to-rtc/udprelay.ts
trevj: Done.

src/socks-to-rtc/udprelay.ts
iislucas: Sorry, I meant "TODO: don't ignore BND.PORT and BND.ADDR ... "

spec/chrome-udpsocket_spec.js
```diff
@@ -9,6 +9,8 @@ describe("chrome-udpsocket", function() {
   var bindResult;
```
iislucas: I think this file should end up as a TS file. add as an issue and a TODO in the file?

src/socks-to-rtc/udprelay.ts
trevj: Good catch. I had a mistake here. It should have referred to the DST.ADDRESS and DST.PORT sent by the SOCKS client during the handshake -- these are typically unspecified, and entirely optional for the server to respect. I clarified the comment. No TODO needed I think.

spec/chrome-udpsocket_spec.js
trevj: We'd have to add this TODO to all our test files, so how about just the issue (good idea BTW):
https://github.com/uProxy/socks-rtc/issues/47

iislucas:
Seems like a good start. merge away.

PR #44, opened 2014-03-07
move to freedom 0.3+

trevj:
This moves socks-rtc to use freedom-runtime-chrome and, transitively, freedom 0.3+.

Mostly mechanical replacement of done() with then() but there were few tricky spots in RtcToNet -- would appreciate any suggestions you have there regardign asynchronicity.

Tested on the CI and locally with grunt endtoend.


trevj:
Oops! Obviously not tested on the CI. Still failing. It's grunt setup, which we don't need anymore. Will followup in a sec.

trevj:
Passing.

Eliminating grunt setup seems to cut our CI test times by about 50% -- hooray!

src/rtc-to-net/netclient.ts
```diff
@@ -89,15 +89,14 @@ module Net {
...
-              'Failed to create socket. createInfo: ' + createInfo));
-        }
-      })
+      return fSockets.create('tcp', {}).then(
+          (createInfo:Sockets.CreateInfo) => {
+            this.socketId = createInfo.socketId;
+            if (!this.socketId) {
+              return Promise.reject(new Error(
+                  'Failed to create socket. createInfo: ' + createInfo));
+            }
```
iislucas, line 98: I think this function should be returning a promise around the `createInfo:Sockets`, no? I think typescript should be complaining here - that probably means the typing around our promise stuff if too lenient.

src/rtc-to-net/netclient.ts
```diff
@@ -89,15 +89,14 @@ module Net {
      * Wrapper which returns a promise for a created socket.
      */
     private createSocket_ = ():Promise<Sockets.CreateInfo> => {
```
iislucas: nit: while we're using no space around argument:argumentType, for return values we are. https://docs.google.com/a/google.com/document/d/1ZV70oTNWO00bs0RcMwqr3FvKJXA6smYQ2kZ8Zr8YZnw/edit#heading=h.gmob2htqhzm4

trevj:
As discussed, let's keep chrome-fsocket.ts around for now. I verified that the register call in boilerplate.js takes precedence over the register call in freedom-runtime-chrome.

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -21,25 +21,35 @@ module RtcToNet {
     private sctpPc:PeerConnection;
     private netClients:{[channelLabel:string]:Net.Client} = {};
 
+    // Static initialiser which returns a promise to create a new Peer
+    // instance complete with a signalling channel.
+    static Create = (peerId:string):Promise<Peer> => {
```
iislucas: Coding style " : " for return values.

src/rtc-to-net/netclient.ts
trevj: Done.

src/rtc-to-net/netclient.ts
trevj, line 98: Good catch. TypeScript doesn't complain if I neglect to return the createInfo...and it does complain if I try to return both Promise.reject and the (raw) createInfo. Ummm, yes we have a problem :-)

https://github.com/uProxy/socks-rtc/issues/45

src/rtc-to-net/rtc-to-net.ts
trevj: Fixed throughout.

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -21,25 +21,35 @@ module RtcToNet {
...
+            });
+          });
+          dbg('signalling channel to SCTP peer connection ready.');
+          F(peer);
+        });
+      });
+    }
+
+    // Don't use this directly -- use Create().
     constructor (public peerId:string) {
```
iislucas: Given that a Peer object doesn't make sense without a created Channel object, I think we should take in the channel object as an argument to the constructor. 

In the Create call above, you then create the peer with the channel object and complete the promise in the then. This way no one can try to create a badly formed promise object by accident.

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -194,8 +203,9 @@ module RtcToNet {
       }
       // TODO: Check for access control?
       // dbg('sending signal to transport: ' + JSON.stringify(signal.data));
-      var peer = this.fetchOrCreatePeer_(signal.peerId);
-      peer.sendSignal(signal.data);
+      this.fetchOrCreatePeer_(signal.peerId).then((peer) => {
+        peer.sendSignal(signal.data);
```
iislucas, line 203: What do you think of a followup patch that renames peer.sendSignal to peer.handleSignal?

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -203,10 +213,12 @@ module RtcToNet {
      */
     private fetchOrCreatePeer_(peerId:string) {
       var peer = this.peers_[peerId];
-      if (peer) {
-        return peer;
+      if (!peer) {
```
iislucas: I think we should avoid checks based on value of an entry unless we really mean the value is there and has a special boolean meaning. Better to use the "x in Y" style.  e.g. 

```
if(!(peerId in this.peers_)) { 
  // create peer
  ... 
  this.peers_[peerId] = ... 
}
return this.peers_[peerId];
```

This gives the code a tad compositionality in the form of copy/paste-ableness too :)

src/socks-to-rtc/tcp.ts
```diff
@@ -97,15 +95,13 @@ module TCP {
...
-                               ' \n Result Code: ' + resultCode);
-        }
-      });
+      return fSockets.listen(this.serverSocketId, this.addr, this.port)
+          .then((resultCode:number) => {  // Ensure the listen was successful.
+            if (0 !== resultCode) {
+              return Util.reject('listen failed on ' + this.endpoint_ +
+                  ' \n Result Code: ' + resultCode);
+            }
+          });
```
iislucas, line 105: Add the return for the success promise with the result code.

iislucas:
Nice work; a couple of small fixes to do.

src/rtc-to-net/rtc-to-net.ts
trevj: Oooh, yes. Much better. Now we can remove the nasty "please do not use this" commen from the constructor and Create() is just a helper method, which I've renamed to CreateWithChannel(). Thanks!

src/rtc-to-net/rtc-to-net.ts
trevj, line 203: I think it would be a good idea. I've added a TODO.

src/rtc-to-net/rtc-to-net.ts
trevj: Yes!

src/socks-to-rtc/tcp.ts
trevj, line 105: Oops. Done. I had to add this though:
return Promise.resolve(resultCode);

otherwise TypeScript complains with:
error TS2198: Could not find the best common type of types of all return statement expressions.

(due to the earlier return of Util.reject)

PR #49, opened 2014-03-11
Cleanups

iislucas:
- Gruntfile now uses (more actively maintained) grunt-typescript instead of slightly grunt-ts
- Gruntfile breaks up copying & typescripting into components
- Gruntfile typescript tasks no longer depends on un-needed ".d" files. 
- Made tests pass in interactive mode (selenium test was previous accidentally tried when it can't run) 
- Made all output files appear in the `build` directory.  Chrome app is in `build/chrome-app`
- Updated the README
- Renamed/changed/commented `boilerplate.js` to `freedom-config.ts`
- Removes accidentally added js file. 

TESTED: `grunt test`


Gruntfile.coffee
```diff
@@ -3,75 +3,62 @@ module.exports = (grunt) ->
...
-      # Eventually, this should be a wildcard once we've figured out how to run
-      # more dependencies under Jasmine.
-      src: [
-        'chrome/js/socks-to-rtc/socks.js',
-        'chrome/js/chrome-udpsocket.js'
-      ],
-      options : {
-        specs : 'spec/**/*_spec.js'
-      }
+      selenium:
```
trevj: I don't think the selenium tests will run with grunt-contrib-jasmine. You should probably remove this.

Gruntfile.coffee
```diff
@@ -3,75 +3,62 @@ module.exports = (grunt) ->
...
+        options : { specs : 'spec/selenium/**/*_spec.js' }
+      socksToRtc:
+        src: ['build/chrome-app/socks-to-rtc/socks.js']
+        options : { specs : 'spec/socks-to-rtc/**/*_spec.js' }
+      chromeProvider:
+        src: ['build/chrome-app/chrome-providers/chrome-udpsocket.js']
+        options : { specs : 'spec/chrome-provider/**/*_spec.js' }
     }
 
     env: {
```
trevj, line 61: I think you need to change this to:
path.resolve('build/chrome-app')

trevj:
Generally I love it. A few comments on grunt endtoend. Just double check that before merging.

Gruntfile.coffee
iislucas: Oh yeah, forgot that was why I split this up to start with... :P

Gruntfile.coffee
iislucas, line 61: yes, thanks!

iislucas:
TESTED: `grunt test` and `grunt endtoend`

PR #53, opened 2014-03-12
refactor SOCKS header decoding and add UDP request decoding

trevj:
This adds support for decoding UDP headers. These are very similar to "regular" SOCKS request headers so I refactored our existing code a little. Also defined TypeScript interfaces for SOCKS requests and UDP requests. I think it's a little better overall.

Next step is probably to have the UDP relay actually relay data.


spec/socks-to-rtc/socks-headers_spec.js
```diff
@@ -0,0 +1,96 @@
+describe("socks", function() {
```
iislucas, line 2: Nice! Maybe add TODOs/issues to add ipv6 requests also.

src/socks-to-rtc/socks.ts
```diff
@@ -325,17 +173,8 @@ module Socks {
     // Given a data |buffer|, interpret the SOCKS request.
     public static interpretRequest = (buffer:ArrayBuffer) => {
```
iislucas: Seems to me like this function is doing very little more now than "Socks.interpretSocksRequest(byteArray, request);" Do we need both? Also: if an exception is raised, should we not convert it to a promise rejection? 

i.e. Make: `interpretSocksRequest` return the `SocksRequest`, and then just call: `Socks.interpretSocksRequest(new Uint8Array(buffer));` ... what do you think?

iislucas:
nice! Just one suggested tweak to have a think about and then feel free to merge.

spec/socks-to-rtc/socks-headers_spec.js
trevj, line 2: AFAICT this file already supports everything we need for IPv6 (which is really just decoding IPv6). Some tests for that would be good, of course, so I added some TODOs to the test.

src/socks-to-rtc/socks.ts
trevj: Yes! Inlined.

Regarding the error, it seems that the next thennable in the chain having a rejection argument receives the error - so in this case the handler at line 160 handles this gracefully. Good point!

PR #54, opened 2014-03-14
Build path update

keroserene:
Puts `socks-to-rtc` and `rtc-to-net` back in the root level of `build/`, so that the user of this repo can use the library without caring about `build/chrome-app`.

Tested: `grunt test` and manual End-to-End

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -10,18 +10,33 @@ module.exports = (grunt) ->
         expand: true, cwd: 'node_modules/freedom-runtime-chrome/'
         src: ['freedom.js']
         dest: 'build/chrome-app/' } ] }
-      chromeApp: { files: [ {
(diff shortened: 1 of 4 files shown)


Gruntfile.coffee
```diff
@@ -10,18 +10,33 @@ module.exports = (grunt) ->
...
-        dest: 'build/chrome-app/' } ] }
+        dest: 'build/' } ] }
       rtc2net: { files: [ {
         expand: true, cwd: 'src/'
         src: ['rtc-to-net/**/*.json']
-        dest: 'build/chrome-app/' } ] }
+        dest: 'build/' } ] }
+
+      chromeApp: { files: [
+        {
```
iislucas: Use condensed style as above.

iislucas:
Looks good, one style nit. merge on!

PR #55, opened 2014-03-17
migrate from peerconnection to transport

trevj:
This moves socks-to-rtc and rtc-to-net over to the new transport API. I ran into some issues with message delivery on Chrome 33 (on Linux) that delayed this by a day.

Two caveats:
- The Transport API provider isn't in freedom-runtime-chrome. So, for right now, I've added Freedom as a runtime dependency just to grab the provider. We can get rid of that as soon as we're ready to move to freedom-for-chrome.
- Peerconnection messages had both text and ArrayBuffer fields -- Transport just has an ArrayBuffer. So, for the "commands" (SOCKS-CONNECT, SOCKS-DISCONNECTED, NET-DISCONNECTED) I've taken to sending them on a special "control" datachannel. Those messages aren't typed, and we should review the general asynchronicity-ness of SOCKS <-> RTC <-> NET events some more (especially since we're now establishing the net connection on its own datachannel) but this seems to work quite well (there were already some TODOs concerning asynchronicity...I'll make this a followup commit).

Tested with grunt test and grunt endtoend (UDP still works, too).

### DIFF
--- a/src/interfaces/communications.d.ts
+++ b/src/interfaces/communications.d.ts
@@ -1,10 +1,18 @@
 // Types for communications between socks-to-rtc and rtc-to-net.
 
 declare module Channel {
+
+  export enum COMMANDS {
+    NET_CONNECT_REQUEST = 1,
+    NET_CONNECT_RESPONSE = 2,
+    NET_DISCONNECTED = 3,
+    SOCKS_DISCONNECTED = 4
+  }
+
   // "Top-level" message for the control channel.
   export interface Command {
     // Name of message, e.g. NetConnectRequest.
-    type:string;
+    type:COMMANDS;
     // Datachannel with which this message is associated.
     tag?:string;
     // JSON-encoded message, e.g. NetConnectRequest.
--- a/src/interfaces/communications.d.ts
+++ b/src/interfaces/communications.d.ts
@@ -1,22 +1,38 @@
 // Types for communications between socks-to-rtc and rtc-to-net.
 
 declare module Channel {
+  // "Top-level" message for the control channel.
+  export interface Command {
+    // Name of message, e.g. NetConnectRequest.
+    type:string;
+    // Datachannel with which this message is associated.
+    tag?:string;
+    // JSON-encoded message, e.g. NetConnectRequest.
+    data?:string;
+  }
 
-  export interface Message {
-    channelLabel:string;
-    text?:string;
-    buffer?:ArrayBuffer;
+  // Sent to request a connection be established with a remote server.
+  export interface NetConnectRequest {
+    // 'tcp' or 'udp'.
+    protocol:string;
+    // Destination address and port.
+    address:string;
+    port:number;
+  }
+
+  export interface NetConnectResponse {
+    // Address and port on which we have made the connection to the
+    // remote server, or both undefined if the connection could not be
+    // made.
+    address?:string;
+    port?:number;
   }
 
   // Should be returned after tieing a data-channel with SOCKS, back to form an
   // endpoint response to the local TCP server.
-  interface EndpointInfo {
+  export interface EndpointInfo {
     ipAddrString:string;
     port:number;
-  }
-
-  interface CloseData {
-    channelId:string;
   }
 
 }  // module Channel
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -104,20 +104,12 @@ module RtcToNet {
     }
 
     /**
-     * Return data from Net to Peer.
-     */
-    private serveDataToPeer_ = (tag:string, data:ArrayBuffer) => {
-      dbg('reply ' + data.byteLength + ' bytes ---> ' + tag);
-      this.transport.send(tag, data);
-    }
-
-    /**
      * Tie a Net.Client for Destination |dest| to data-channel |tag|.
      */
     private prepareNetChannelLifecycle_ =
         (tag:string, dest:Net.Destination) => {
       var netClient = this.netClients[tag] = new Net.Client(
-          (data) => { this.serveDataToPeer_(tag, data); },  // onResponse
+          (data) => { this.transport.send(tag, data); },  // onResponse
           dest);
       // Send NetClient remote disconnections back to SOCKS peer, then shut the
       // data channel locally.
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -93,23 +93,26 @@ module RtcToNet {
             host: request.address,
             port: request.port
           };
-          // This is what we'll send to the client.
-          // If we successfully connect to the remote host then we'll
-          // populate the address and port fields.
-          var response:Channel.NetConnectResponse = {};
           this.prepareNetChannelLifecycle_(command.tag, dest)
               .then((endpointInfo:Channel.EndpointInfo) => {
-                response.address = endpointInfo.ipAddrString;
-                response.port = endpointInfo.port;
+                return endpointInfo;
+              }, (e) => {
+                dbgWarn('could not create netclient: ' + e.message);
+                return undefined;
               })
-              .then(() => {
+              .then((endpointInfo?:Channel.EndpointInfo) => {
+                var response:Channel.NetConnectResponse = {};
+                if (endpointInfo) {
+                  response.address = endpointInfo.ipAddrString;
+                  response.port = endpointInfo.port;
+                }
                 var out:Channel.Command = {
                     type: 'NetConnectResponse',
                     tag: command.tag,
                     data: JSON.stringify(response)
                 }
-                this.transport.send('control', ArrayBuffers.stringToArrayBuffer(
-                    JSON.stringify(out)));
+                this.transport.send('control',
+                    ArrayBuffers.stringToArrayBuffer(JSON.stringify(out)));
               });
         } else {
           // TODO: support SocksDisconnected command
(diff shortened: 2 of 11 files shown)


src/rtc-to-net/rtc-to-net.ts
```diff
@@ -66,89 +66,72 @@ module RtcToNet {
...
-        this.prepareNetChannelLifecycle_(label, dest);
 
-      } else if (message.buffer) {
-        dbg(label + ' <--- received ' + JSON.stringify(message));
-        if(!(label in this.netClients)) {
-          dbgErr('[RtcToNet] non-existent channel! Msg: ' +
-              JSON.stringify(message));
+      if (message.tag == 'control') {
+        var commandText = ArrayBuffers.arrayBufferToString(message.data);
+        var command:any = JSON.parse(commandText);
```
iislucas: avoid "any" typing. "Object" is probably the right thing here; or even better give it a typing that you expect. :)

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -66,89 +66,72 @@ module RtcToNet {
...
+        if (command.command == 'SOCKS-CONNECT') {
+          // Text from the peer indicates request for a new destination.
+          // Assumes |message.text| is a Net.Destination.
```
iislucas: broken comment?

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -66,89 +66,72 @@ module RtcToNet {
...
+          if (command.tag in this.netClients) {
+            // TODO: This shouldn't be fired! This is bad!
```
iislucas: not clear what the TODO is for now. I think we're warning appropriately... drop the TODO?

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -66,89 +66,72 @@ module RtcToNet {
...
+            dbgWarn('Net.Client already exists for datachannel: ' + command.tag);
+            return;
+          }
+          var dest:Net.Destination = JSON.parse(commandText);
```
iislucas: didn't we just parse commandText above?

I think making explicit types for the command would help understand the code here.

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -66,89 +66,72 @@ module RtcToNet {
...
      */
-    private serveDataToPeer_ = (channelLabel:string, data:ArrayBuffer) => {
-      // TODO: peer connection is firing a response for *every* channelLabel.
-      // This needs to be fixed.
-      dbg('reply ' + data.byteLength + ' bytes ---> ' + channelLabel);
-      this.sctpPc.send({
-          'channelLabel': channelLabel,
-          'buffer': data
-      });
+    private serveDataToPeer_ = (tag:string, data:ArrayBuffer) => {
```
iislucas: We probably don't need this private function if we just call this.transport.send everywhere serveDataToPeer_ was being called... although maybe this makes code more readable. Your call.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -104,137 +104,103 @@ module SocksToRTC {
...
-    private replyToSOCKS_ = (msg:Channel.Message) => {
-      var label = msg.channelLabel;
-      if (!label) {
-        dbgErr('received message without channelLabel! msg: ' +
-            JSON.stringify(msg));
-        return;
-      }
-      if (!(label in this.socksSessions)) {
-        dbgErr(label + ' not associated with SOCKS session!');
+    private replyToSOCKS_ = (msg:freedom.Transport.IncomingMessage) => {
```
iislucas: I don't think I find the name replyToSocks to be helpful... it makes me think this is part of the SOCKS handshake. This is really about passing on data from the peer... what do you think of calling it: `onDataFromPeer` ?

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -104,137 +104,103 @@ module SocksToRTC {
...
-      if (!(label in this.socksSessions)) {
-        // This can happen if both peers send disconnection signals at the same
-        // time.
-        dbgWarn('No SOCKs session to close for ' + label);
-        return;
-      }
-      // End SOCKS session.
-      this.socksServer.endSession(this.socksSessions[label]);
-      delete this.socksSessions[label];
+    private closeConnection_ = (tag:string) => {
```
iislucas: How about calling this `closeConnectionToPeer`. Just to be clear it's the connection to a single peer that is being closed.

src/rtc-to-net/rtc-to-net.ts
trevj: I was thinking about this a little more. What do you think about adding types to src/interfaces/communications.d.ts to define these 'control' messages?

src/rtc-to-net/rtc-to-net.ts
trevj: Ah, yes. Thanks. Removed.

src/rtc-to-net/rtc-to-net.ts
trevj: Done.

src/rtc-to-net/rtc-to-net.ts
trevj: See previous comment -- perhaps we can define these messages in src/interfaces/communications.d.ts? Once done, I can clean up this section.

src/rtc-to-net/rtc-to-net.ts
trevj: Well spotted. It's only called in one place so I just inlined it.

src/socks-to-rtc/socks-to-rtc.ts
trevj: Agreed! Renamed.

src/socks-to-rtc/socks-to-rtc.ts
trevj: Done.

src/rtc-to-net/rtc-to-net.ts
iislucas: Sounds good.

trevj:
Some updates, pretty much as discussed:
- define types for the messages flowing around the datachannel
- when we send a SOCKS-CONNECT message, we wait for a reply from the remote server before sending a response to the client (previously, we assumed the connect would succeed)

The second part is a little tricky, would appreciate any feedback you have (socks-to-rtc.ts).

As a side effect, we're now finally piping the endpointInfo through from netclient.ts all the way to the SOCKS client. Yay!

Unit and Selenium tests passing.

trevj:
So, I still need to rebuild the freedom-typescript-api NPM to pull in the transport API declaration but otherwise I think this is ready...

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -66,117 +66,97 @@ module RtcToNet {
...
+            port: request.port
+          };
+          // This is what we'll send to the client.
+          // If we successfully connect to the remote host then we'll
+          // populate the address and port fields.
+          var response:Channel.NetConnectResponse = {};
+          this.prepareNetChannelLifecycle_(command.tag, dest)
+              .then((endpointInfo:Channel.EndpointInfo) => {
+                response.address = endpointInfo.ipAddrString;
+                response.port = endpointInfo.port;
```
keroserene: Maybe we could make the promise-chaining clearer by creating and returning `response` here, and receiving a `:Channel.NetConnectResponse` as the argument in the next `then(...)`. This way we'll avoid mutating something declared outside of the asynchronous context, which could become risky.

src/interfaces/communications.d.ts
```diff
@@ -1,24 +1,40 @@
 // Types for communications between socks-to-rtc and rtc-to-net.
 
 declare module Channel {
+  // "Top-level" message for the control channel.
+  export interface Command {
+    // Name of message, e.g. NetConnectRequest.
+    type:string;
```
keroserene: Maybe we could make this an Enum? It'd be a more robust typecheck, and it'll go from full strings to just an integer on the wire :)

src/rtc-to-net/rtc-to-net.ts
keroserene: I put that TODO in there for the assumption that we weren't reusing datachannels, and also that there is only one request per `NetClient`. I think that's still currently the case. If this warning is fired, and we aren't using a data channel pool, then something really bad bight be going on from the `socks-to-rtc` side?

src/rtc-to-net/rtc-to-net.ts
keroserene: Oh, one of my new comments is exactly this. An enum in communications.d.ts sounds great!

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -104,137 +108,130 @@ module SocksToRTC {
...
-    private replyToSOCKS_ = (msg:Channel.Message) => {
-      var label = msg.channelLabel;
-      if (!label) {
-        dbgErr('received message without channelLabel! msg: ' +
-            JSON.stringify(msg));
-        return;
-      }
-      if (!(label in this.socksSessions)) {
-        dbgErr(label + ' not associated with SOCKS session!');
+    private onDataFromPeer = (msg:freedom.Transport.IncomingMessage) => {
```
keroserene: nit: underscore suffix for private methods :)

keroserene:
Looks good! Just a couple things and then it should totally merge. :)

src/rtc-to-net/rtc-to-net.ts
trevj: Yes! I still haven't fully gotten the hang of chaining results from promises together but this is much better. Thanks.

src/interfaces/communications.d.ts
trevj: Yes! Thanks. Never thought of that.

src/socks-to-rtc/socks-to-rtc.ts
trevj: Oops, I've been ignoring this all over the place. Fixed throughout.

PR #58, opened 2014-03-23
use uproxy-build-tools & use typescript tests

iislucas:
- Moved third party into build tools (it's was just the jasmine DefinitelyTyped jasmine interfaces)
- Added npm-dependency for build-tools
- Made tests into typescript files
- Moved tests to be next to what they test. 
- Used github's Apache2 LICENSE file

TESTED: grunt test


trevj:
Nice. I much prefer co-locating the tests and code.

PR #59, opened 2014-03-26
Pipe UDP traffic through the encrypted data channel

trevj:
This completes UDP support.

Main change here is how socks.ts and socks-to-rtc.ts communicate: Socks.Session is no more, and now they both use EndpointInfo objects to talk to one another. Socks.Session wasn't working for me because one UDP session can result in connections to many different hosts (meaning multiple data channels). A new class, Socks.UdpSession, handles this complication: it decodes the UDP headers, keeping udprelay.ts and, on the remote side, udpclient.ts, nice and simple (in fact they're almost identical and some refactoring in the future could unify them).

Existing unit and integration tests pass. UDP support tested with the help of redsocks and an echo server:
https://docs.google.com/a/google.com/document/d/150cwwDqfTe0_Os-FTTVO9DNG4wqxNlksxxo2QeFj1bA/edit?usp=sharing

Note: I can't guarantee this will support, say, a Skype call but I think if it doesn't it should be a matter of bug fixes rather than a fundamental redesign.


src/rtc-to-net/rtc-to-net.ts
```diff
@@ -85,16 +90,13 @@ module RtcToNet {
...
             return;
           }
-          var dest:Net.Destination = {
-            host: request.address,
-            port: request.port
-          };
-          this.prepareNetChannelLifecycle_(command.tag, dest)
-              .then((endpointInfo:Channel.EndpointInfo) => {
+          this.prepareNetChannelLifecycle_(command.tag, request)
+              .then((endpointInfo:Net.Endpoint) => {
```
keroserene: Looks like you've updated `prepareNetChannelLifecycle_` to return a promise for an `Net.Endpoint` rather than an `EndpointInfo` - rename this variable to just `endpoint`?

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -120,13 +122,15 @@ module RtcToNet {
         }
       } else {
         dbg(message.tag + ' <--- received ' + JSON.stringify(message));
-        if(!(message.tag in this.netClients)) {
+        if(message.tag in this.netClients) {
+          dbg('forwarding ' + message.data.byteLength + ' bytes from datachannel ' + message.tag);
+          this.netClients[message.tag].send(message.data);
+        } else if (message.tag in this.udpClients) {
+          dbg('forwarding ' + message.data.byteLength + ' bytes from datachannel ' + message.tag);
```
keroserene: nit: Two duplicate `dbg` statements is the beginning of a code smell. Maybe conditionally set the 'index' for `this.udpClients` or `this.netClients`, and then you'll only have to call `send` once. Otherwise, indicate in the `dbg` message whether its TCP or UDP so it's not exactly the same :)

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -148,21 +144,16 @@ module SocksToRTC {
...
-      session.onRecv((buf) => { this.sendToPeer_(tag, buf); });
-      session.onceDisconnected().then(() => {
-        var command:Channel.Command = {
-            type: Channel.COMMANDS.SOCKS_DISCONNECTED,
-            tag: tag
-        };
-        this.transport_.send('control', ArrayBuffers.stringToArrayBuffer(
-            JSON.stringify(command)));
-      });
+    private terminate_ = (tag:string) => {
```
keroserene, line 151: Should there be an existence check for `tag`?

src/socks-to-rtc/socks.ts
```diff
@@ -23,95 +24,154 @@ module Socks {
...
-          .then(Socks.Session.getHandshake)
-          .then(Socks.Session.checkVersion)
-          .then(Socks.Session.checkAuth)
-          // Success - Create new session.
+    private establishSession_ = (conn:TCP.Connection): void => {
+      var socksRequest:SocksRequest = {};
+      var udpRelay:Socks.UdpRelay;
+      conn.receive()
+          .then((buffer:ArrayBuffer) => {
+            Server.validateHandshake(buffer);
```
keroserene: This can simplify to `.then(Server.validateHandshake)`

src/socks-to-rtc/socks.ts
```diff
@@ -23,95 +24,154 @@ module Socks {
...
-    public static checkVersion(handshakeBytes:Uint8Array) {
+    static validateHandshake(buffer:ArrayBuffer) {
+      var handshakeBytes = new Uint8Array(buffer);
+
+      // Only SOCKS Version 5 is supported.
       var socksVersion = handshakeBytes[0];
-      if (Socks.VERSION5 != socksVersion) {
-        return Util.reject('unsupported version: ' + socksVersion);
+      if (socksVersion != Socks.VERSION5) {
+        throw new Error('unsupported SOCKS version: ' + socksVersion);
```
keroserene, line 170: I think that `throw` doesn't get caught by the promise fulfillment/rejection chain. We should stick to `Promise.reject` so that `establishSession_` above can catch it correctly.

src/socks-to-rtc/socks.ts
```diff
@@ -23,95 +24,154 @@ module Socks {
...
      */
-    public static checkVersion(handshakeBytes:Uint8Array) {
+    static validateHandshake(buffer:ArrayBuffer) {
+      var handshakeBytes = new Uint8Array(buffer);
+
+      // Only SOCKS Version 5 is supported.
       var socksVersion = handshakeBytes[0];
-      if (Socks.VERSION5 != socksVersion) {
-        return Util.reject('unsupported version: ' + socksVersion);
+      if (socksVersion != Socks.VERSION5) {
```
keroserene, line 169: Why are we reversing the yoda conditional? (I've been seeing this throughout...)
It's better to have the constant on the left-hand side, because this eliminates all chance of subtle bugs of accidental assignment if perchance one types `=` rather than `==`. Also, it's more consistent, because the most important info on the conditional comes first while reading code)

src/socks-to-rtc/socks.ts
```diff
@@ -211,48 +223,91 @@ module Socks {
...
+      var header = new ArrayBuffer(headerLength);
+      var headerBytes = new Uint8Array(header);
+      for (var i = 0; i < header.byteLength; i++) {
+        headerBytes[i] = bytes[i];
+      }
+      var payload = new ArrayBuffer(bytes.byteLength - headerLength);
+      var payloadBytes = new Uint8Array(payload);
+      for (var i = 0; i < payload.byteLength; i++) {
+        payloadBytes[i] = bytes[headerLength + i];
+      }
```
keroserene: This could be simplified to `payloadBytes = bytes.slice(headerLength, headerLength + payload.byteLength)`

src/socks-to-rtc/socks.ts
```diff
@@ -211,48 +223,91 @@ module Socks {
...
+    private onData_ = (data:ArrayBuffer) => {
+      // Split the datagram into two parts: the UDP header and the payload.
+      // TODO: have interpretUdpRequest return an integer which we can use here
+      var headerLength = 10;
+      var bytes = new Uint8Array(data);
+      var header = new ArrayBuffer(headerLength);
+      var headerBytes = new Uint8Array(header);
+      for (var i = 0; i < header.byteLength; i++) {
+        headerBytes[i] = bytes[i];
+      }
```
keroserene: `headerBytes = bytes.slice(0, header.byteLength)`

src/socks-to-rtc/udprelay.ts
```diff
@@ -103,11 +103,36 @@ module Socks {
     }
 
     private onSocksClientData = (recvFromInfo:UdpSocket.RecvFromInfo) => {
```
keroserene: nit: underscore suffix

keroserene:
Looks totally reasonable, just a bunch of nits :)

src/rtc-to-net/rtc-to-net.ts
trevj: Good catch. Thanks!

src/rtc-to-net/rtc-to-net.ts
trevj: Ah...quite right :-) I would like, as a followup, to have a common superclass for TCP and UDP clients -- then it will be easy to remove this near-duplication.

src/socks-to-rtc/socks-to-rtc.ts
trevj, line 151: Yes! And in closeConnectionToPeer. And terminate_ should also discard the reference to tag. Thanks!

src/socks-to-rtc/socks.ts
trevj: Done.

src/socks-to-rtc/socks.ts
trevj, line 170: Done.

src/socks-to-rtc/socks.ts
trevj, line 169: I've never been a fan of yoda conditions. I'm a little disappointed to discover just now that there's a vague justification for using them in JavaScript but I find the reversed word order much harder to read, which for outweighs the tiny chance of typo-ing '=' with '==' (my theory: if you can't express some code in words then you don't really understand it, and reversing the word order hurts this process).

src/socks-to-rtc/socks.ts
trevj: Right. Unfortunately, I don't know how to tell TypeScript that ArrayBuffer.slice is a thing (it's unsupported by phantomjs, too, for our unit tests).

src/socks-to-rtc/socks.ts
trevj: see previous comment

src/socks-to-rtc/socks.ts
keroserene: Oh, hmm. Well, I don't know about phantomjs, but for typescript, every `interface` stanza extends existing interfaces if they exist. (I've had to do this to get around weird `window` things... so it should work if you put this in one of the `.d.ts` files:

```
interface ArrayBuffer {
  slice: (begin:number, end?:number) => ArrayBuffer
}
```

trevj: Ah, thanks! I added this to arraybuffer.d.ts. Unfortunately it doesn't gain us a whole lot since copying still has to be done byte by byte...but this is probably better for now.

src/socks-to-rtc/socks.ts
```diff
@@ -211,48 +222,82 @@ module Socks {
...
-    /**
-     * Send |buffer| to session's TCP client.
-     */
-    public sendData = (buffer) => { this.tcpConnection.sendRaw(buffer); }
+    private onData_ = (data:ArrayBuffer) => {
+      // Split the datagram into two parts: the UDP header and the payload.
+      // TODO: have interpretUdpRequest return an integer which we can use here
+      var headerLength = 10;
+      var header = data.slice(0, headerLength);
+      var headerBytes = new Uint8Array(header);
```
iislucas, line 251: You don't need to copy bytes unless you want to change them and preserve the old copy. see: Uint8Array constructor that can start from an offset and have a defined length: 

Uint8Array Uint8Array(ArrayBuffer buffer, optional unsigned long byteOffset, optional unsigned long length);

PR #62, opened 2014-03-27
Convert socks-rtc to use new core.tcpsocket API

dborkan:
- Convert socks-rtc to use new core.tcpsocket API
- Remove unused methods from tcp.ts
- Change freedom library from "freedom-runtime-chrome" to "freedom-for-chrome"
- Tested with echo-server and performed end-to-end test using "curl -x socks5h://localhost:9999 www.google.com"
- Will create a separate pull request to update typescript API at node_modules/freedom-typescript-api/interfaces/tcp-socket.d.ts

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -22,6 +22,10 @@ module.exports = (grunt) ->
         expand: true, cwd: 'src/'
         src: ['rtc-to-net/**/*.json']
         dest: 'build/chrome-app/' } ] }
+      echo: { files: [ {
+        expand: true, cwd: 'test/'
+        src: ['*.js']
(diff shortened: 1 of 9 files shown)


trevj:
Great!

I think the echo server might make more sense in freedom-for-chrome, basically to test the TCP provider -- what do you think?

Also, please do submit the update for freedom-typescript-api ASAP! Right now there's a compile error which interestingly doesn't cause our continuous build to fail...I filed an issue:
https://github.com/uProxy/socks-rtc/issues/63

dborkan:
OK, also that issue is similar to this one I created for the UProxy repo:
https://github.com/UWNetworksLab/UProxy/issues/187

PR #64, opened 2014-03-27
Fix TypeScript errors related to conversion to new core.tcpsocket.

dborkan:
Fix TypeScript errors related to conversion to new core.tcpsocket.

This doesn't yet fix the following warning with Promise.all:

> > /Users/dborkan/github/socks-rtc/src/socks-to-rtc/tcp.ts(150,17):
> > error TS2082: Supplied parameters do not match any signature of call target:
> >  Type 'any[]' is missing property 'then' from type 'Thenable<{}>'.


iislucas:
LGTM

PR #65, opened 2014-03-28
fail on typescript errors and remove base_path option warning

trevj:
This will make the build Grunt task fail if TypeScript finds any type errors (the default is to fail only if TypeScript can't create an output JS file). Also rename base_path -> basePath, to remove annoying compiler warnings.

I'll wait until master is back in good shape to merge this.

Addresses:
https://github.com/uProxy/socks-rtc/issues/63


iislucas:
LGTM! Just need to merge in changes from upstream with `git pull origin master`, check that all things work, then you can merge. 

Thanks for doing this!

PR #66, opened 2014-03-28
fix issues caused during previous merge

trevj:


trevj:
So, I took commit 94c578862cd4f9e6c2029e96c7401effa5e258d6 and cherry picked what I believe to be the relevant commits after that point. Then, I rebased against master. As you can see, the diff turned out to be quite small (mostly some issues with the freedom manifests).

PR #71, opened 2014-04-03
remove the custom udp provider

trevj:
This makes socks-rtc use the UDP provider in freedom-for-chome. Also update to 0.1.4 of freedom-for-chrome which has the recent getInfo fix and test:
https://github.com/freedomjs/freedom-for-chrome/pull/6

npm also seemed to no longer like the version of grunt-typescript we were specifiying...the new version here seems to work just fine.


trevj:
Ooh, node 0.8 doesn't like how some of the packages are specified now in package.json:
https://travis-ci.org/uProxy/socks-rtc/builds/22148535

Do we care about node 0.8?

iislucas:
I don't care about node 0.8 :-)

Ooh, node 0.8 doesn't like how some of the packages are specified now in
package.json:
https://travis-ci.org/uProxy/socks-rtc/builds/22148535

Do we care about node 0.8?

—
Reply to this email directly or view it on
GitHubhttps://github.com/uProxy/socks-rtc/pull/71#issuecomment-39399973
.

PR #72, opened 2014-04-18
SVG-icise the Travis badge and add a badge to track NPM dependencies

trevj:
Not bothering with runtime dependencies because I think we probably should make our single runtime dependency into a dev dependency.


iislucas:
LGTM

PR #74, opened 2014-04-18
make freedom runtime dep into a dev dep so its not our only runtime dep

trevj:
This made sense to me at the time...but now it's our only runtime dependency and there's no reason really why it's any different to any of our other dependencies.


iislucas:
LGTM

PR #75, opened 2014-04-18
Create firefox build of socks-rtc

Paul-E:
This just gets the echo server working. I tested it by running the firefox-app extension in FF (instructions how in README) and connecting to the echo server with `telnet localhost 9998`. Text sent through telnet was echoed back.


iislucas:
Cool! Please add in this (github) comment log how this was tested (so that we know what to do to run/test the echo server in FF)

Paul-E:
There is info in the README on how to run the extension in FF. See  959e951

Gruntfile.coffee
```diff
@@ -113,12 +142,16 @@ module.exports = (grunt) ->
     'typescript:rtc2net'
     'typescript:common'
     'typescript:chromeApp'
-    'copy:freedom'
-    'copy:freedomProviders'
+    'copy:freedomChrome'
```
trevj, line 151: No need to do this right now but I think it might be useful to add separate "grunt chrome" and "grunt firefox" targets. Looks like the changes here will support that.

src/firefox-app/data/socks_rtc.json
```diff
@@ -0,0 +1,17 @@
+{
```
trevj: So, it would be really nice not to have to duplicate this file and socks_to_rtc_to_net.js -- they look the same. If it isn't too hard, I wonder could the Gruntfile just copy these from chrome-app into firefox-app? (at some point in the future, we could think of a better common location).

trevj:
Nice! Just one suggestion for preventing duplicate files. I think you need to merge in some recent changes and then we're good to go!

trevj:
And apologies for not getting to this yesterday!

Paul-E:
I got rid of the duplicate files and now copy them over from the chrome app directory instead. I agree, eventually the builds could be refactored a bit.

iislucas:
Please add an issue for it; and add a comment with the issue number to this
thread.  Thanks!

PR #77, opened 2014-05-02
use common utilities from uproxy-build-tools

trevj:
This is how socks-rtc would use the utilities proposed in:
https://github.com/uProxy/build-tools/pull/1


iislucas:
LGTM. But lets get uProxy/build-tools#1 in first, then make sure Travis is happy, then submit this. :)

PR #79, opened 2014-05-07
Add support for proxying by using newer freedom-for-firefox.

Paul-E:
Fixes #78


iislucas:
This was a small enough change not to really matter; but lets try and make sure we keep the habit of 1. saying how changes are tested and 2. having someone review them and gives the :+1:   :)

PR #81, opened 2014-05-08
Move code around for inclusion in UProxy.

dborkan:
Move code around for inclusion in UProxy.
- Moved tcp_echo_server to its own freedom module, so that socks-to-rtc.json doesn't include the echo server (i.e. UProxy shouldn't include the echo server)
- Moved util and freedom-providers into build directory

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -145,6 +145,8 @@ module.exports = (grunt) ->
   # Define the tasks
   taskManager = new TaskManager.Manager();
 
+  # TODO: create separate build commands for just the socks-to-rtc and rtc-to-net
+  # libaries, chrome app, and firefox app.
   taskManager.add 'build', [
     'typescript'
     'copy'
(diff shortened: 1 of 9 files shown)


Gruntfile.coffee
```diff
@@ -137,19 +146,8 @@ module.exports = (grunt) ->
...
-    'copy:freedomProvidersChrome'
-    'copy:freedomProvidersFirefox'
-    'copy:socks2rtc'
-    'copy:rtc2net'
-    'copy:echoChrome'
-    'copy:echoFirefox'
-    'copy:chromeApp'
-    'copy:firefoxApp'
+    'typescript'
+    'copy'
```
keroserene, line 152: Cool :) One question: what happened to the granular builds? For instance:
- Just the socks2rtc and rtc2net modules (for deployment as a library on npm)
- just the chrome example app
- just the firefox example app

By default, socks-rtc should only build itself as a library.

iislucas, line 152: Better to create several tasks using taskManager, one for each thing you might want to build. Don't worry, repeated mentions of a sub-task will never be done more than once. But it does give finer level granularity for what you want to build as well as better watch rules.

iislucas:
nice cleanup. Either add a TODO to properly structure the tasks & follow up later; or do it as part of this one. :) 

:+1:

Gruntfile.coffee
dborkan, line 152: just added a TODO for separating the build rules

PR #84, opened 2014-05-29
Emit socksToRtcSuccess or socksToRtcFailure after freedom.transport.setup succeeds or fails.

dborkan:
Emit socksToRtcSuccess or socksToRtcFailure after freedom.transport.setup succeeds or fails.

Tested with socks-rtc chrome app to verify that these signals fire.

### DIFF
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -38,7 +38,11 @@ module RtcToNet {
       this.transport = freedom['transport']();
       this.transport.on('onData', this.passPeerDataToNet_);
       this.transport.on('onClose', this.closeNetClient_);
-      this.transport.setup('RtcToNet-' + peerId, channel.identifier);
+      this.transport.setup('RtcToNet-' + peerId, channel.identifier).then(
+        // TODO: emit signals when peer-to-peer connections are setup or fail.
+        () => { console.log('RtcToNet transport.setup succeeded'); },
+        () => { console.error('RtcToNet transport.setup failed'); }
+      );
       this.signallingChannel = channel.channel;
       this.signallingChannel.on('message', (msg) => {
         freedom.emit('sendSignalToPeer', {
@@ -116,9 +120,13 @@ module RtcToNet {
                 this.transport.send('control',
                     ArrayBuffers.stringToArrayBuffer(JSON.stringify(out)));
               });
+        } else if (command.type === Channel.COMMANDS.HELLO) {
+          // Hello command is used to establish communication from socks-to-rtc,
+          // just ignore it.
+          dbg('received hello.');
         } else {
           // TODO: support SocksDisconnected command
-          dbgWarn('unsupported control command: ' + command.type);
+          dbgWarn('unsupported control command: ' + JSON.stringify(command));
         }
       } else {
         dbg(message.tag + ' <--- received ' + JSON.stringify(message));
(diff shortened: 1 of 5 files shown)


src/rtc-to-net/rtc-to-net.ts
```diff
@@ -116,9 +120,13 @@ module RtcToNet {
                 this.transport.send('control',
                     ArrayBuffers.stringToArrayBuffer(JSON.stringify(out)));
               });
+        } else if (command.type === Channel.COMMANDS.HELLO) {
+          // Hello command is used to establish communication from socks-to-rtc,
+          // just ignore it.
+          dbg('received hello.');
```
keroserene, line 126: Just a question to keep my understanding up-to-date: Is the HELLO command what actually indicates the data channel is ready? How is this related to the `transport.setup` callbacks which are TODO for emitting success and failure signals on the peer connection? (I'm assuming something more will be required here when we want to accurately indicate a peer is providing access, along with bandwidth information)

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -55,7 +59,17 @@ module SocksToRTC {
       // Messages received via signalling channel must reach the remote peer
       // through something other than the peerconnection. (e.g. XMPP)
       fCore.createChannel().then((chan) => {
-        this.transport_.setup('SocksToRtc-' + peerId, chan.identifier);
+        this.transport_.setup('SocksToRtc-' + peerId, chan.identifier).then(
+          () => {
+            console.log('SocksToRtc transport_.setup succeeded');
```
keroserene: Should the `console.log` and `console.error`s be `dbg*`s? (This way all log messages can be disabled later, if we want)

keroserene:
> Tested with socks-rtc chrome app to verify that these signals fire.

Do you have a modification to the js in the chrome app which installs `freedom.on('socksToRtcSuccess')` and `freedom.on('socksToRtcFailure')` handlers, to confirm that pathway goes all the way through?

src/rtc-to-net/rtc-to-net.ts
dborkan, line 126: The HELLO command is what results in the first data channel being opened through WebRTC, which will cause freedom.transport.setup's returned promise to fulfill.  I had added the initiateConnection option to peerconnection.setup (code here https://github.com/freedomjs/freedom/pull/47/files#diff-32d48f939f7c111639df5eaae044650bR428) but after discussing with Will we decided not to expose this to callers of freedom.transport.setup (because in theory users of freedom.transport shouldn't need to worry about how the connection is initiated or by which side).  The HELLO message is really just a hack for now, until a longer term fix that will probably involve both sides of a peer-connection negotiating using a nonce to decide who should initiate the connection, followed by sending an initial message across to fulfill the freedom.transport.setup promise

keroserene, line 126: Ah, cool. Do the `freedom.transport.setup` promises on both the socks-to-rtc and rtc-to-net sides current fulfill behind the scenes in freedom, as soon as the first data channel is fully ready (through this HELLO message hack)? If so, maybe a comment/TODO describing this may be useful (unless there's already a comment somewhere else)

dborkan, line 126: Yes, that's correct.  I just updated the comment here and added a TODO.  I believe the freedom team is going to be making changes for this on issue https://github.com/freedomjs/freedom/issues/1 (although the title of that issue is no longer accurate).

dborkan:
Thanks for the review.   I just updated src/chrome-app/socks_to_rtc_to_net.js to print out trace when those signals are received, which appears after the line "[SocksToRtc] SocksToRtc transport_.setup succeeded"

keroserene:
Hurray! :+1:

PR #88, opened 2014-06-05
Emit signals from rtc-to-net when a client starts/stops proxying.

dborkan:
Emit signals from rtc-to-net when a client starts/stops proxying.

Tested manually by including in uProxy, and "grunt test".


iislucas:
:+1:

PR #89, opened 2014-06-06
Add ping-pong (heartbeat) to detect if 1 side of a peer connection closes

dborkan:
Add ping-pong (heartbeat) to detect if 1 side of a peer connection closes.  This fixes issue https://github.com/uProxy/uProxy/issues/151 (so that if a client closes, the server's UI can remove the X is proxying through you text).  This is also necessary to have in place for fixing issue https://github.com/uProxy/uProxy/issues/178 (that will require some new UI)

Tested with manually and with "grunt test"


src/rtc-to-net/rtc-to-net.ts
```diff
@@ -37,10 +41,13 @@ module RtcToNet {
...
       this.transport = freedom['transport']();
       this.transport.on('onData', this.passPeerDataToNet_);
-      this.transport.on('onClose', this.closeNetClient_);
+      this.transport.on('onClose', this.onCloseHandler_);
       this.transport.setup('RtcToNet-' + peerId, channel.identifier).then(
         // TODO: emit signals when peer-to-peer connections are setup or fail.
-        () => { dbg('RtcToNet transport.setup succeeded'); },
+        () => {
+          dbg('RtcToNet transport.setup succeeded');
+          this.startPingPong_();;
```
trevj: typo (two semi-colons)

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -217,6 +242,10 @@ module SocksToRTC {
           // Receiving a disconnect on the remote peer should close SOCKS.
           dbg(command.tag + ' <--- received NET-DISCONNECTED');
           this.closeConnectionToPeer(command.tag);
+        } else if (command.type === Channel.COMMANDS.PONG) {
+          // Receiving a disconnect on the remote peer should close SOCKS.
```
trevj: I think this comment can be removed :-)

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -68,14 +75,16 @@ module RtcToNet {
     /**
      * Close PeerConnection and all TCP sockets.
      */
-    public close = () => {
+    private onCloseHandler_ = () => {
+      dbg('transport closed from peerId ' + this.peerId);
```
trevj: Perhaps re-think the wording here? It's not clear to me what "transport closed from x" means!

trevj:
Cool.

Some nits and two broader things to consider:
1. It would be great to have a comment explaining why this is necessary because, as i understand it, WebRTC should have its own ping mechanism (to keep NAT mappings alive, if nothing else). For example, there is a comment explaining the 'start' message which also explains that the 'start' message will eventually go away.
2. If each side will consider the connection a failure unless it receives a ping from the other within some time interval then it seems there's no real need for a ping-pong cycle? If so, the code becomes basically the same on both sides and can probably be pulled out into some common location.

FWIW, I think #1 is more important than #2 :-)

dborkan:
Just added some comments about why we need ping-pong (the short answer is webrtc doesn't notify us if the peer disconnects).

Also the code isn't shared because it is different on the client (socks-to-rtc) and server (rtc-to-net) sides.  For example on the client side we need 2 timers: 1 to send pings every second, another to check for pongs every 10 seconds.  But on the server side, we only need 1 timer to check for pings every 10 seconds, because the pongs are sent back immediately after a ping is received.

PR #90, opened 2014-06-08
Emit signal when ping-pong times out on socks-to-rtc side

dborkan:
Emit signal when ping-pong times out on socks-to-rtc side

Tested by merging into uProxy and verifying that we get a signal when the server disconnects, also re-ran "grunt test"


iislucas:
:+1:

PR #91, opened 2014-06-08
Remove peer from server after closing, so that a new rtc-to-net connection can be started later with the same peerId

dborkan:
Prior to this change, if Alice proxied through Bob, then closed the proxy, then restarted a proxy through Bob, rtc-to-net on Bob's (the server) side would fail because it would try to re-use an already closed Peer (wrapper around freedom.transport).  I'm fixing this by having the Peer remove itself from the server's mapping after it is closed, that way a new Peer object will be created each time a proxy is attempted.

This is needed for issue https://github.com/uProxy/uProxy/issues/44 (proxying a 2nd time with the same peer fails), although it does not completely fix that issue (we are still hitting XMPP 503 (probably throttling) errors the 2nd time, for which we have issue https://github.com/uProxy/uProxy/issues/186)

Tested manually (verified we no longer see errors about setRemoteDescription failing for a closed RTCPeerConnection objected), and re-ran grunt test.

### DIFF
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -23,6 +23,7 @@ module RtcToNet {
     // TODO: this is messy...a common superclass would help
     private netClients:{[tag:string]:Net.Client} = {};
     private udpClients:{[tag:string]:Net.UdpClient} = {};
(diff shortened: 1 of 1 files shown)


iislucas:
:+1:

PR #92, opened 2014-06-09
batch signalling channel messages and send them at most every second

trevj:
Please don't review this just yet...the code isn't tidy or tested. Just for experimentation.

### DIFF
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -53,12 +53,33 @@ module RtcToNet {
         },
         (e) => { dbgErr('RtcToNet transport.setup failed ' + e); }
       );
+
+      // signalling channel messages are batched and dispatched each second.
+      // the unshift/pop business is supposed to help prevent clobbering messages.
+      // TODO: kill this loop!
+      // TODO: size limit on batched message
+      // TODO: this code is completely common to socks-to-rtc (growing need for shared lib)
+      var queuedMessages = [];
+      setInterval(() => {
+        var batch = [];
+        while (queuedMessages.length > 0) {
+          batch.push(queuedMessages.pop());
+        }
+        if (batch.length > 0) {
+          dbg('dispatching signalling channel messages...');
+          freedom.emit('sendSignalToPeer', {
+            peerId: peerId,
+            data: batch.join('===')
+          });
+        }
+      }, 1000);
+
       this.signallingChannel = channel.channel;
       this.signallingChannel.on('message', (msg) => {
-        freedom.emit('sendSignalToPeer', {
-            peerId: peerId,
-            data: msg
-        });
+        // msg is plain-text, e.g.:
+        //   {"candidate":{"sdpMLineIndex":0,"sdpMid":"audio","candidate":"a=candidate:1587603213 1 udp 2122260223 69.91.136.43 57727 typ host generation 0\r\n"}} 
+        dbg('signalling channel message: ' + msg);
+        queuedMessages.unshift(msg);
       });
       dbg('signalling channel to SCTP peer connection ready.');
     }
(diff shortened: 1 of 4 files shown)


trevj:
Oh, two important things to explain:
1. This doesn't just batch ICE candidate messages, it batches _all_ messages. The benefit is that we don't need to parse the messages. It seems to work just fine.
2. I'm using '===' as a message separator. Yes, that's weird but candidate lines normally contain newlines so I just picked something that's very unlikely to appear in an actual message.

matthild:
If it's possible for messages to contain the separator, then some kind of escaping or delimiting is necessary to guarantee correctness, right?

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -53,12 +53,33 @@ module RtcToNet {
...
       this.signallingChannel = channel.channel;
       this.signallingChannel.on('message', (msg) => {
-        freedom.emit('sendSignalToPeer', {
-            peerId: peerId,
-            data: msg
-        });
+        // msg is plain-text, e.g.:
+        //   {"candidate":{"sdpMLineIndex":0,"sdpMid":"audio","candidate":"a=candidate:1587603213 1 udp 2122260223 69.91.136.43 57727 typ host generation 0\r\n"}} 
+        dbg('signalling channel message: ' + msg);
+        queuedMessages.unshift(msg);
```
dborkan: Why unshift?  It seems like you'd want to push the messages to the end of the queue rather than treat it like a stack

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -53,12 +53,33 @@ module RtcToNet {
...
+      // signalling channel messages are batched and dispatched each second.
+      // the unshift/pop business is supposed to help prevent clobbering messages.
+      // TODO: kill this loop!
+      // TODO: size limit on batched message
+      // TODO: this code is completely common to socks-to-rtc (growing need for shared lib)
+      var queuedMessages = [];
+      setInterval(() => {
+        var batch = [];
+        while (queuedMessages.length > 0) {
+          batch.push(queuedMessages.pop());
```
dborkan: Is copying messages from queuedMessages to batch necessary?

src/rtc-to-net/rtc-to-net.ts
trevj: pop() removes from the end, push() appends to the end -- this is my primitive attempt to help prevent the dispatcher from clobbering new incoming messages.

src/rtc-to-net/rtc-to-net.ts
trevj: no but it makes it easier to run join() :-)

(and they're only references -- so the array is pretty lightweight)

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -53,12 +53,33 @@ module RtcToNet {
...
+        }
+        if (batch.length > 0) {
+          dbg('dispatching signalling channel messages...');
+          freedom.emit('sendSignalToPeer', {
+            peerId: peerId,
+            data: batch.join('===')
```
dborkan: Maybe use 4 = signs, like ====, since === is actually valid javascript (not that anyone really should be passing messages containing javascript, but they might)

dborkan:
Awesome news, I just tested this in uProxy, and I was able to start, stop, and restart a proxy with the same peer within 1 minute and it worked!  It also appeared that what had been 10 XMPP messages were consolidated into 2 messages.

After lunch I will try even faster starting/stopping/starting/etc and see how fast we can go before getting 503 errors from GTalk

keroserene:
Excellent! :clap: 

Another suggestion along this path: Would it be possible to procure actual rate-limiting constants specific to the various XMPP networks we have providers for, and prematurely rate-limit messages on uProxy's side (with maybe even some additional margin) accordingly? That way we can be much closer to avoiding being throttled (although not absolutely guaranteed) / waiting for a penalty beyond our control to lift.

trevj:
So, I updated this to use JSON objects with the following prototype:
{
  version: number;
  messages :string[]
}

Will throw an error unless a version 1 object is received.

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -273,7 +291,20 @@ module RtcToNet {
       // TODO: Check for access control?
       // dbg('sending signal to transport: ' + JSON.stringify(signal.data));
       this.fetchOrCreatePeer_(signal.peerId).then((peer) => {
-        peer.sendSignal(signal.data);
+        // TODO: this code is completely common to rtc-to-net (growing need for shared lib)
+        try {
+          var batchedMessages :Channel.BatchedMessages = JSON.parse(signal.data);
```
matthild, line 296: At some point we should also check that parsing succeeded and that the resulting JS object is well-formed (i.e., that it has the properties we expect and that those properties' values have the correct types).  I think it's fine to leave all that as a TODO for now, but before long it would be worthwhile to validate this over-the-wire input before consuming it.

PR #93, opened 2014-06-11
Fixed cleanup of on-connection callbacks.

matthild:
Remove the callback entry from the collection, rather than keeping the entry in the collection but setting its callback to 'undefined'.


iislucas:
:+1:

PR #94, opened 2014-06-11
Big Networking Cleanup

iislucas:
There is still a bit more work to do in a followup CL to remove the EndpointInfo interface - there's bugs there. Also on RtcToNet to properly encapsulate each session. 

Big changes: 
- TCP connection isn't just for connections we receive: it can be used to create new connections too
- Removed NetClient, this is now just a tcp connection
- New simplified build directory layout has a typescript subdirectory for all typescript code to have a common path structure to find dependent files. 
- Uses new build-tools structure. 
- freedom files now called freedom.json
- Cleaned up/standardized use of Net.Endpoint (represents both an IP address and a port)


src/tcp/tcp.ts
```diff
@@ -0,0 +1,326 @@
...
+
+  function endpointOfSocketInfo(info:TcpSocket.SocketInfo) : Net.Endpoint {
+     return { address: info.peerAddress, port: info.peerPort }
+  }
+
+  // A limit on the max number of TCP connections before we start rejecting
+  // new ones.
+  var DEFAULT_MAX_CONNECTIONS = 1048576;
+
+  // Tcp.Server: a TCP Server. This listens for connections when listen is
```
trevj, line 22: You've converted the comments in this file to C++ style -- is there a reason for that? We should probably stick to one or the other, at least for class and method-level comments.

src/tcp/tcp.ts
```diff
@@ -0,0 +1,326 @@
...
+  // called, and handles the new connection as specified by the onConnection
+  // argument to the constructor.
+  export class Server {
```
trevj, line 25: So, in general I'm ambivalent about this class because it doesn't provide any value besides a slightly nicer API than Freedom TCP sockets (and that's very subjective). Did you consider just dropping it altogether?

src/tcp/tcp.ts
```diff
@@ -0,0 +1,326 @@
...
+    private conns:{[socketId:number] : Tcp.Connection} = {};
+
+    // Create TCP server.
+    // `address` = Address to be listening on.
+    // `port` = the port to listen on; 0 = dynamic allocation.
+    // `onConnection` = the handler for new TCP Connections.
+    // `maxConnections` = the number of connections after which all new ones
+    // will be closed as soon as they connect.
+    // TODO: make address and port into getters; we don't support changing them.
+    constructor(public address        :string,
```
trevj, line 37: I think onConnection and maxConnections can be private, as can address and port if we add getter methods for them.

src/tcp/tcp.ts
```diff
@@ -0,0 +1,326 @@
...
+        // Stop too many connections.  We create a new socket here from the
+        // incoming Id and immediately close it, because we don't yet have a
+        // reference to the incomming socket.
+        freedom['core.tcpsocket'](socketId).close();
+        dbgErr('Too many connections: ' + connectionsCount);
+        return;
+      }
+
+      // if we don't know how to handle the connection, so close it.
+      if (!this.onConnection) {
```
trevj, line 72: When would this ever happen? I think it would be better to have the constructor fail if the server is created with a null connection handler (seems like the server is useless without it).

package.json
```diff
@@ -17,10 +18,10 @@
     "grunt-jasmine-node": "^0.2.1",
     "webdriverjs": "^1.6.1",
     "grunt-env": "^0.4.1",
-    "uproxy-build-tools": "~0.0.5",
+    "uproxy-build-tools": "^1.0.0",
     "freedom-for-chrome": "^0.1.4",
     "freedom-for-firefox": "^0.1.1",
-    "freedom-typescript-api": "~0.1.5",
+    "freedom-typescript-api": "^0.2.0",
```
trevj, line 24: The latest published version is 1.0.0 -- did you forget to re-publish?:
npm info freedom-typescript-api

package.json
```diff
@@ -17,10 +18,10 @@
     "grunt-jasmine-node": "^0.2.1",
     "webdriverjs": "^1.6.1",
     "grunt-env": "^0.4.1",
-    "uproxy-build-tools": "~0.0.5",
+    "uproxy-build-tools": "^1.0.0",
```
trevj, line 21: The latest published version is 0.0.5 -- did you forget to re-publish?:
npm info uproxy-build-tools

src/tcp/tcp.ts
```diff
@@ -0,0 +1,326 @@
...
+        return;
+      }
+
+      // Create new connection.
+      dbg('Tcp.Server accepted connection on socket id ' + socketId);
+      var conn = new Connection({existingSocketId:socketId});
+      // When the connection is disconnected correctly, or by error, remove
+      // from the server's list of connections.
+      conn.onceDisconnected.then(
+        () => {
```
trevj, line 84: So, there's some weirdness with the Freedom TCP socket provider in that even when a connection is closed _normally_, it results in an "error". This is easy to observe with the help of the echo server:
telnet localhost 9998
hello
hello
^]quit

(error message on javascript debug console)

trevj, line 84: I filed an issue for this against freedom-for-chrome:
https://github.com/freedomjs/freedom-for-chrome/issues/12

src/tcp/tcp.ts
```diff
@@ -0,0 +1,326 @@
...
+        dbgErr('No connection handler is defined!');
+        return;
+      }
+
+      // Create new connection.
+      dbg('Tcp.Server accepted connection on socket id ' + socketId);
+      var conn = new Connection({existingSocketId:socketId});
+      // When the connection is disconnected correctly, or by error, remove
+      // from the server's list of connections.
+      conn.onceDisconnected.then(
```
trevj, line 83: I _think_ you can reduce this to:

conn.onceDisconnected
    .then(() => { dbg('yay!); }, (e) => { dbgWarn('boo'); })
    .then(() => { delete this.conns[socketId]; });

Right?

src/tcp/tcp.ts
```diff
@@ -0,0 +1,326 @@
...
+    }
+  }  // class Tcp.Server
+
+  /**
+   * Tcp.Connection - Wraps up a single TCP connection to a client
+   *
+   * @param {number} socketId The ID of the server<->client socket.
+   */
+  export class Connection {
+    // Unique identifier for each connection.
```
trevj, line 138: Unique to the server which created this Connection instance?

src/tcp/tcp.ts
```diff
@@ -0,0 +1,326 @@
...
+
+      // Close all Tcp connections.
+      for (var i in this.conns) {
+        var c = this.conns[i];
+        allPromises.push(c.close());
+      }
+
+      // Wait for all promises to complete.
+      return Promise.all(allPromises).then(() => {
+        dbg('successfully closed all Tcp Connections.');
```
trevj, line 127: Please add an error handler which prints a warning if this fails.

src/tcp/tcp.ts
```diff
@@ -0,0 +1,326 @@
...
+      });
+    }
+  }  // class Tcp.Server
+
+  /**
+   * Tcp.Connection - Wraps up a single TCP connection to a client
+   *
+   * @param {number} socketId The ID of the server<->client socket.
+   */
+  export class Connection {
```
trevj, line 137: So, in general this class still seems way over-engineered to me. For example, why bother queuing data until a read handler is specified rather than requiring clients to add a handler immediately?

This is related to my previous comment regarding how much we need this TCP wrapper class at all.

src/chrome-app/endtoend.spec.ts
```diff
@@ -1,4 +1,7 @@
-/// <reference path='../../node_modules/uproxy-build-tools/third_party/DefinitelyTyped/jasmine/jasmine.d.ts' />
+// This is a selenium demo end-to-end spec. But it currently doesn't actually
+// test anything we care about.
```
trevj, line 2: It tests that proxying works at all, which is arguably something we care about.

src/chrome-app/socks-to-rtc-to-net.ts
```diff
@@ -0,0 +1,70 @@
...
+  export function SocksToRtc() : freedom.SocksToRtc;
+  export function RtcToNet() : freedom.RtcToNet;
+}
+
+var LOCALHOST = '127.0.0.1';
+var DEFAULT_ECHO_PORT = 9998;
+var DEFAULT_SOCKS_PORT = 9999;
+var fakeSessionId = 'some fake session id';
+
+// CONSIDER: When networking code stabalises, we could remove the echo server.
```
trevj, line 27: I quite like the coinage but I would just make this a TODO because while some IDEs will flag TODO but none that I know of flag CONSIDER :-)

src/socks-to-rtc/freedom.json
```diff
@@ -3,12 +3,13 @@
   "description": "Listen for incoming requests, and relay them through a transport.",
```
trevj, line 3: So, I re-opened the thread on this namign convention:
https://groups.google.com/forum/#!topic/uproxy-eng/x9bjdpMG4eQ

Short version: the manifest's filename is incorporated into the webworker name, and using a common manifest filename currently makes the webworkers impossible to identify by name alone.

src/udp/udpclient.ts
```diff
@@ -1,6 +1,7 @@
-/// <reference path='../../node_modules/freedom-typescript-api/interfaces/freedom.d.ts' />
-/// <reference path='../../node_modules/freedom-typescript-api/interfaces/udp-socket.d.ts' />
-/// <reference path='../../node_modules/freedom-typescript-api/interfaces/promise.d.ts' />
+/// <reference path='../freedom-typescript-api/interfaces/freedom.d.ts' />
```
trevj, line 1: So, having the new tcp/ and udp/ folders as siblings imply that they logically related but they're not: tcp.ts is a general purpose TCP server whereas the UDP classes are only useful to the SOCKS server. I would move this folder under socks/.

src/socks-to-rtc/freedom.json
trevj, line 3: Seems like Freedom 0.5 will solve this by incorporating the module's name instead. Disregard.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -1,89 +1,84 @@
...
-     * Start the Peer, based on the remote peer's info.
+     * Start the SocksToRtc Peer, based on the the local host and port to
+     * listen on.
      * This will emit a socksToRtcSuccess signal when the peer connection is esablished,
      * or a socksToRtcFailure signal if there is an error openeing the peer connection.
      * TODO: update this to return a promise that fulfills/rejects, after freedom v0.5
      * is ready.
      */
-    public start = (remotePeer:PeerInfo) => {
+    public start = (endpoint:Net.Endpoint) => {
```
trevj, line 56: +1 to this change.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -1,89 +1,84 @@
...
+      // the Freedom channel object to send messages to the peer.
       fCore.createChannel().then((chan) => {
-        this.transport_.setup('SocksToRtc-' + peerId, chan.identifier).then(
+        this.transport_.setup('SocksToRtc', chan.identifier).then(
           () => {
             dbg('SocksToRtc transport_.setup succeeded');
-            freedom.emit('socksToRtcSuccess', remotePeer);
-            this.startPingPong_();
+            freedom.emit('socksToRtcSuccess', endpoint);
+            // this.startPingPong_();
```
trevj, line 71: Did you forget to uncomment this?

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -1,89 +1,84 @@
...
       this.reset_();  // Begin with fresh components.
-      dbg('starting - target peer: ' + JSON.stringify(remotePeer));
-      // Bind peerID to scope so promise can work.
-      this.remotePeer_ = remotePeer;
-      var peerId = this.peerId_ = remotePeer.peerId;
-      if (!peerId) {
-        dbgErr('no Peer ID provided! cannot connect.');
-        return false;
-      }
+      dbg('starting SocksToRtc: ' + JSON.stringify(endpoint));
```
trevj, line 58: Suggest making this something like 'starting on xxxx:yyyy' because dbg() will prefix with SocksToRtc anyway.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -358,28 +349,24 @@ module SocksToRTC {
...
 
   var modulePrefix_ = '[SocksToRtc] ';
   function dbg(msg:string) { console.log(modulePrefix_ + msg); }
   function dbgWarn(msg:string) { console.warn(modulePrefix_ + msg); }
   function dbgErr(msg:string) { console.error(modulePrefix_ + msg); }
 
-}  // module SocksToRTC
+}  // module SocksToRtc
 
+// This is what is avauilable to Freedom.
```
trevj, line 362: Typo, but I'm also not sure what the comment means.

src/socks-to-rtc/socks.ts
```diff
@@ -5,46 +5,54 @@
...
   export class Server {
 
-    private tcpServer:TCP.Server;
+    private tcpServer:Tcp.Server;
 
     /**
      * @param address local interface on which to bind the server
      * @param port port on which to bind the server
-     * @param createChannel_ function to create a new datachannel
+     * @param createDataChannel_ function to create a new datachannel
```
trevj, line 34: The args descriptions in the comment need to be refreshed.

iislucas:
This has become https://github.com/uProxy/uproxy-networking/pull/101

PR #97, opened 2014-06-27
added tsd files for webrtc and es6-promises

iislucas:
Started using tsd to import DefinitelyTyped files.


iislucas:
This is now part of https://github.com/uProxy/uproxy-networking/pull/101

PR #100, opened 2014-07-09
move to Freedom v0.5-based freedom-for-xxx providers

trevj:
This moves to Freedom v0.5-based `freedom-for-chrome` and `freedom-for-firefox`. No minor version bump since this doesn't require an API change. My next task is to give `socks-rtc` a custom API; that would obviously be a breaking change and should bump us to v0.2.

Tested with `grunt endtoend` and with curl in both Firefox and Chrome.

Only issue I see are some error messages in Chrome:

```
```

I wasn't able to quickly track down the source of these. Since my next task is to give `socks-rtc` a custom API, I intend to track down the issue then. All seems to work fine; my guess is this error reporting is new in Freedom 0.5.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -13,7 +13,7 @@ module.exports = (grunt) ->
         src: ['freedom-for-chrome.js']
         dest: 'build/chrome-app/' } ] }
       freedomFirefox: { files: [ {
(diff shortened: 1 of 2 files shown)


dborkan:
:thumbsup:

PR #103, opened 2014-08-27
Restrict access to standard global IP addresses.

bemasc:


bemasc:
This is a fix for https://github.com/uProxy/uproxy/issues/253

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -299,6 +317,20 @@ module RtcToNet {
...
 
+    private isAllowedAddress_ = (addressString:string) : boolean => {
+      // ipaddr.process automatically converts IPv4-mapped IPv6 addresses into
+      // IPv4 Address objects.  This ensure that an attacker cannot reach a
+      // restricted IPv4 endpoint that is identified by its IPv6-mapped address.
+      try {
+        var address = ipaddr.process(addressString);
+        return address.range() == 'unicast';
+      } catch (e) {
+        log.error(this.longId() + ': error in isAllowedAddress_(' +
```
iislucas: Not sure we want to error here. I'd leave just the error at the `handleWebRtcControlMessage_`

iislucas:
Looks great, just one comment that we probably don't need to double-report error. Otherwise :+1: Fix and merge on!

src/rtc-to-net/rtc-to-net.ts
bemasc: Done.

PR #107, opened 2014-09-05
updated deps

iislucas:
Updated deps w.r.t. Travis telling us what was out of date. 
TESTED: 
Ran sample apps. (had to go in and build node_modules/uproxy-lib by hand; will have a followup to fix that)


trevj:
:+1:

PR #109, opened 2014-09-05
require.resolve to avoid node_module paths, and fix postinstall script

iislucas:
- remove references to node_module paths by using require.resolve
- added a post-npm install script 

TESTED: ran & tested from clean clone:
`npm install`
then open sample app socks-rtc-net app and use it to do some proxying.


keroserene:
:+1:

PR #111, opened 2014-09-08
fixed package deps; fixed .DS_store ignore

iislucas:


keroserene:
:+1: (as soon as now un-used grunt rules are removed)

PR #113, opened 2014-09-09
add a SOCKS demo with a copy/paste-based signalling channel

trevj:
Also rename the parent folder to `socks-server`, in line with `echo-server`.


Gruntfile.coffee
```diff
@@ -224,33 +223,31 @@ module.exports = (grunt) ->
     'copy:echoServerChromeApp'
   ]
 
-  #-------------------------------------------------------------------------
-  # tasks for sample apps
-  taskManager.add 'socksRtcNet', [
+  taskManager.add 'socksSamples', [
```
iislucas, line 231: Maybe good to call it socksServerSamples not to confuse it with something that might be in the socks folder?

trevj, line 231: What do you think of renaming the `socks` folder to `socks-common`, and keeping this as `socksSamples`?

trevj:
Oops, looks like Github only supports one assignee.

Gruntfile.coffee
iislucas, line 231: that works.

trevj, line 231: OK, I just renamed it to `socks-common`.

matthild:
:+1:

PR #114, opened 2014-09-09
fix the echo server, and move it under tcp/ module

trevj:
This fixes the echo server and moves it under `tcp/`, since that is the component under test.

Also moves it to port 9998 in case the developer is also running the SOCKS sample, which uses port 9999.


trevj:
Depends on:
https://github.com/uProxy/uproxy-networking/pull/113

iislucas:
:+1: 

(assuming it was tested manually)

PR #117, opened 2014-09-11
use churn rather than uproxypeerconnection

trevj:
This integrates the SOCKS server with churn. Pretty simple, thanks to the identical interfaces. Depends on:
https://github.com/uProxy/uproxy-churn/pull/40

I wish there were a simple way to plug in different peerconnection providers but that seems like something to address at some future point.

### DIFF
--- a/src/socks-server/samples/simple-socks-chromeapp/freedom-module.json
+++ b/src/socks-server/samples/simple-socks-chromeapp/freedom-module.json
@@ -1,6 +1,6 @@
 {
   "name": "simple-socks",
-  "description": "Single page SOCKS server.",
+  "description": "Single-page SOCKS server.",
   "app": {
     "script": [
       "lib/arraybuffers/arraybuffers.js",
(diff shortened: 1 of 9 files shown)


src/socks-server/samples/simple-socks-chromeapp/freedom-module.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "socks-rtc-net",
-  "description": "Sample socks over rtc to net Freedom manifest",
+  "name": "simple-socks",
+  "description": "Single page SOCKS server.",
```
matthild: "Single page" -> "Single-page"

Also, Github says the Travis build failed, presumably that's because https://github.com/uProxy/uproxy-churn/pull/40 wasn't merged yet at the time.  Before merging, if you haven't already done so, please check that the build passes.

Otherwise :+1:

trevj: Thanks Matt! I fixed that description, re-published `uproxy-churn`, and now see that Travis built successfully.

PR #119, opened 2014-09-12
added stop-promise to socks-rtc-net

iislucas:
- Added promise support for when socks-rtc stops (e.g. peer-connection breaks or is closed by remote peer).
- Added declaration files (*.d) for socks-to-rtc and rtc-to-net
- Made typescript sym-links done by file, not directory (improves support for `cd` then `tsc` testing). 
- rtc-to-net now has 'close' as public
- churn is now used by semantic versioning

### DIFF
--- a/src/socks-to-rtc/socks-to-rtc.ts
+++ b/src/socks-to-rtc/socks-to-rtc.ts
@@ -261,7 +261,7 @@ module SocksToRtc {
       return this.onceClosed;
     }
 
-    public handleDataFromPeer(data:WebRtc.Data) {
+    public handleDataFromPeer = (data:WebRtc.Data) : void => {
       this.dataFromPeer_.handle(data);
     }
 
--- a/src/socks-to-rtc/socks-to-rtc.ts
+++ b/src/socks-to-rtc/socks-to-rtc.ts
@@ -33,7 +33,8 @@ module SocksToRtc {
 
     private isStopped_ :boolean;
     public isStopped = () : boolean => { return this.isStopped_; }
-    public onceStopped :Promise<void>;
+    public onceStopped_ :Promise<void>;
+    public onceStopped = () : Promise<void> => { return this.onceStopped_; }
 
     // Message handler queues to/from the peer.
     public signalsForPeer :Handler.Queue<WebRtc.SignallingMessage, void>;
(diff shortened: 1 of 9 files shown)


src/tcp/tcp.ts
```diff
@@ -163,11 +163,13 @@ module Tcp {
     public stopListening = () : Promise<void> => {
       // Close the server socket.
       return this.serverSocket_.close().then(() => {
-        log.debug('successfully stopped listening for more connections.');
-      });
+          log.debug('successfully stopped listening for more connections.');
+        });
```
keroserene: Why 2 indents? Aren't the other promise chains of 1 indent?

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -26,9 +26,15 @@ module SocksToRtc {
...
   // The |SocksToRtc| class runs a SOCKS5 proxy server which passes requests
   // remotely through WebRTC peer connections.
+  // TODO: rename this 'Server'.
   export class SocksToRtc {
     // Holds the IP/port that the localhost socks server is listeneing to.
     public onceReady : Promise<Net.Endpoint>;
+
+    private isStopped_ :boolean;
+    public isStopped = () : boolean => { return this.isStopped_; }
+    public onceStopped :Promise<void>;
```
keroserene: Should make the internal promise (`private onceStopped_`), and then expose a `public onceStopped()` method. This prevents any chance of the user of `uproxy-networking` accidentally overwriting the promise itself, making it useless. (e.g Potentially doing something like `socksToRtc.onceStopped = someFunction`) It would be more clear if it must always be `socksToRtc.onceStopped().then( ... )

src/tcp/tcp.ts
iislucas: meh. ok. :)

src/socks-to-rtc/socks-to-rtc.ts
iislucas: Yeah, you're right. feels like typescript should supported read-only vars in its interfaces... but you're right, we should use a function for now. Right now this is being done for consistency with other libraries. I'll submit a pull request to change them all at once, but I'll do that separately.

iislucas: clarification: fixed this one in this pull request; will fix the others in a separate pull request.

iislucas:
@keroserene  PTAL

iislucas:
Note: build will be broken until https://github.com/uProxy/uproxy-lib/pull/68 lands.

keroserene:
Yay! :+1:

PR #120, opened 2014-09-12
add parameter to enable/disable obfuscation

trevj:
A quickie, to unblock Matt's improved copy and paste app and to facilitate debugging on Firefox.

This adds a parameter to enable obfuscation, rather than forcing it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-networking",
   "description": "uProxy's networking library: SOCKS5 over WebRTC",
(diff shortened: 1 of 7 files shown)


trevj:
This is also a non-breaking change, which will probably be appreciated while we're still trying to put uProxy back together.

trevj:
Longer term, we should create some kind of config object which `socks-to-rtc` and `rtc-to-net` can accept.

matthild:
:+1:

Thanks!

PR #121, opened 2014-09-12
Fix description of what the copy/paste sample app does.

matthild:
- It does not actually obfuscate yet.
- Removed redundant occurrence of the word "manual".


trevj:
:+1:

PR #122, opened 2014-09-15
update README for new sample apps

trevj:


trevj:
Thanks to Ben for the suggestion "split proxy"!

PR #123, opened 2014-09-18
disable churn in simple-socks for now

trevj:
I can't make >1 request via the proxy with churn. Disabling for now so Lally can continue benchmarking.


mollyling:
LGTM

PR #130, opened 2014-09-24
Emit number of bytes received/sent and display in copypaste sample

lucyhe:
Ran grunt test.

### DIFF
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -37,6 +37,16 @@ module RtcToNet {
     public signalsForPeer :Handler.Queue<WebRtc.SignallingMessage, void> =
         new Handler.Queue<WebRtc.SignallingMessage,void>();
 
+    // For the two Queues below, only bytes sent and received via ArrayBuffers
+    // are considered. Bytes in strings are not considered because of potential
+    // variations in their character to byte translation. (This includes channel
+    // labels, which are sent as strings.)
+    // Queue of the number of bytes received from the peer. Handler is typically
+    // defined in the class that creates an instance of SocksToRtc.
+    public bytesReceivedFromPeer :Handler.Queue<number, void>;
+    // Queue of the number of bytes sent to the peer. Handler is typically
+    // defined in the class that creates an instance of SocktsToRtc.
+    public bytesSentToPeer :Handler.Queue<number,void>;
     // This promise is fulfilled once the peer connection is stablished and
     // this module is ready to start making tcp connections.
     public onceReady :Promise<void>;
@@ -71,6 +81,8 @@ module RtcToNet {
       // through something other than the peerconnection. (e.g. XMPP). This is
       // the Freedom channel object to sends signalling messages to the peer.
       // SOCKS sessions biject to peerconnection datachannels.
+      this.bytesReceivedFromPeer = new Handler.Queue<number, void>();
+      this.bytesSentToPeer = new Handler.Queue<number, void>();
       this.sessions_ = {};
       this.proxyConfig = proxyConfig;
       this.peerConnection_ = obfuscate ?
@@ -87,7 +99,7 @@ module RtcToNet {
           log.debug('dummy init channel.');
           return;
         }
-        var session = new Session(this.peerConnection_, channelLabel, proxyConfig);
+        var session = new Session(this.peerConnection_, channelLabel, proxyConfig, this.bytesReceivedFromPeer, this.bytesSentToPeer);
         this.sessions_[channelLabel] = session;
         session.onceClosed.then(() => {
           delete this.sessions_[channelLabel];
@@ -104,6 +116,8 @@ module RtcToNet {
     // associated TCP connections. Note: once closed, cannot be openned again.
     public close = () => {
       this.peerConnection_.close();
+      this.bytesReceivedFromPeer.clear();
+      this.bytesSentToPeer.clear();
       // CONSIDER: will peerConnection's closing of channels make this un-
       // needed? is it better to include this anyway?
       var channelLabel :string;
@@ -131,6 +145,12 @@ module RtcToNet {
       if(!(rtcData.channelLabel in this.sessions_)) {
         log.error('onDataFromPeer_: no such channel to send data to: ' +
             rtcData.channelLabel);
+        // We only log data sent via ArrayBuffer, so just check .data (and not
+        // .str).
+        if(rtcData.message.buffer) {
+          this.bytesReceivedFromPeer
+            .handle(rtcData.message.buffer.byteLength);
+        }
         return;
       }
       this.sessions_[rtcData.channelLabel].handleWebRtcDataFromPeer(
@@ -188,7 +208,9 @@ module RtcToNet {
         private peerConnection_:WebrtcLib.Pc,
         // The channel Label is a unique id for this data channel and session.
         private channelLabel_:string,
-        public proxyConfig :ProxyConfig) {
+        public proxyConfig :ProxyConfig,
+        private bytesReceivedFromPeer:Handler.Queue<number,void>,
+        private bytesSentToPeer:Handler.Queue<number,void>) {
       this.proxyConfig = proxyConfig;
       this.isClosed_ = false;
       this.hasConnectedToEndpoint_ = false;
@@ -232,6 +254,7 @@ module RtcToNet {
         if (!this.hasConnectedToEndpoint_) {
           log.error(this.longId() + ': Client attempted to send data to ' +
               'tcp connection before it was opened');
+          this.bytesReceivedFromPeer.handle(webrtcData.buffer.byteLength);
           return;
         }
         log.debug(this.longId() + ': passing on data from pc connection to tcp (' +
@@ -239,6 +262,7 @@ module RtcToNet {
         // Note: tcpConnection is smart: it buffers and only sends when it is
         // ready.
         this.tcpConnection.dataToSocketQueue.handle(webrtcData.buffer);
+        this.bytesReceivedFromPeer.handle(webrtcData.buffer.byteLength);
       } else {
         log.error(this.longId() + ': handleWebRtcDataFromPeer: Bad rtcData: ' +
             JSON.stringify(webrtcData));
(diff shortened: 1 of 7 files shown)


src/rtc-to-net/rtc-to-net.ts
```diff
@@ -37,6 +37,16 @@ module RtcToNet {
...
+    // For the two Queues below, only bytes sent and received via ArrayBuffers
+    // are considered. Bytes in strings are not considered because of potential
+    // variations in their character to byte translation. (This includes channel
+    // labels, which are sent as strings.)
+    // Queue of the number of bytes received from the peer. Handler is typically
+    // defined in the class that creates an instance of SocksToRtc.
+    public bytesReceivedFromPeer :Handler.Queue<number, void>;
+    // Queue of the number of bytes sent to the peer. Handler is typically
+    // defined in the class that creates an instance of SocktsToRtc.
+    public bytesSentToPeer :Handler.Queue<number,void>;
```
keroserene: nit: space after `<number,` :)

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -87,7 +99,7 @@ module RtcToNet {
           log.debug('dummy init channel.');
           return;
         }
-        var session = new Session(this.peerConnection_, channelLabel, proxyConfig);
+        var session = new Session(this.peerConnection_, channelLabel, proxyConfig, this.bytesReceivedFromPeer, this.bytesSentToPeer);
```
keroserene: We should stick to the 80-char width limit, at least for the typescript files

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -131,6 +145,12 @@ module RtcToNet {
       if(!(rtcData.channelLabel in this.sessions_)) {
         log.error('onDataFromPeer_: no such channel to send data to: ' +
             rtcData.channelLabel);
+        // We only log data sent via ArrayBuffer, so just check .data (and not
```
keroserene: Did you mean .buffer here? :)

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -71,6 +81,8 @@ module RtcToNet {
       // through something other than the peerconnection. (e.g. XMPP). This is
       // the Freedom channel object to sends signalling messages to the peer.
       // SOCKS sessions biject to peerconnection datachannels.
```
keroserene, line 88: Here's some comment suggestions :)

Let's put this longer comment above, about the signalling channel lower, right above the code it's talking about.

Then, maybe a remark about how the `Handler.Queue` for bytes sent and received is shared and pushed to by each session.

keroserene:
Looking good! Just some nitpickin suggestions

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -37,6 +37,16 @@ module RtcToNet {
     public signalsForPeer :Handler.Queue<WebRtc.SignallingMessage, void> =
         new Handler.Queue<WebRtc.SignallingMessage,void>();
 
+    // For the two Queues below, only bytes sent and received via ArrayBuffers
+    // are considered. Bytes in strings are not considered because of potential
+    // variations in their character to byte translation. (This includes channel
+    // labels, which are sent as strings.)
+    // Queue of the number of bytes received from the peer. Handler is typically
+    // defined in the class that creates an instance of SocksToRtc.
+    public bytesReceivedFromPeer :Handler.Queue<number, void>;
```
trevj: So, I think the most precise explanation is that these queues report the number of bytes transferred between the SOCKS client and the remote host(s) with which it wishes to connect. WebRTC overhead (DTLS headers, ICE initiation, etc.) is not included (because WebRTC does not provide easy access to that data) nor is SOCKS protocol-related data (because it's sent via string messages).

FWIW, once we have obfuscation fully integrated we _will_ be able to report the exact numbers -- but for now this is a good approximation considering that we can't easily extract the true numbers via the WebRTC library.

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -71,6 +81,8 @@ module RtcToNet {
       // through something other than the peerconnection. (e.g. XMPP). This is
       // the Freedom channel object to sends signalling messages to the peer.
       // SOCKS sessions biject to peerconnection datachannels.
+      this.bytesReceivedFromPeer = new Handler.Queue<number, void>();
```
trevj: I think you could create the queues where these variables are declared, and save a couple of lines (just a suggestion: obviously this is fine so it's whatever you prefer).

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -104,6 +116,8 @@ module RtcToNet {
     // associated TCP connections. Note: once closed, cannot be openned again.
     public close = () => {
       this.peerConnection_.close();
+      this.bytesReceivedFromPeer.clear();
```
trevj: Since the `RtcToNet` object can't be re-used I don't think there's any benefit to clearing the byte count queues.

src/rtc-to-net/rtc-to-net.ts
trevj: It looks like this will only report the byte count if the channel label is unknown -- probably not what we want :-)

Looks like we should move this outside of the if block so that we can report the byte count whether or not the channel is known to us.

(apologies if I'm totally wrong on this but that's what it looks like at this late, bleary-eyed hour)

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -232,6 +254,7 @@ module RtcToNet {
         if (!this.hasConnectedToEndpoint_) {
           log.error(this.longId() + ': Client attempted to send data to ' +
               'tcp connection before it was opened');
+          this.bytesReceivedFromPeer.handle(webrtcData.buffer.byteLength);
```
trevj: So, it looks like this function should only report bytes received _once_ and it should do so whether or not the endpoint is yet connected.

src/socks-server/samples/copypaste-socks-chromeapp/freedom-module.ts
```diff
@@ -63,6 +63,17 @@ freedom.on('start', () => {
     freedom.emit('signalForPeer', signal);
   });
 
+  // SocksToRtc adds the number of bytes it sends/receives to its respective
+  // queue as it proxies. When new numbers (of bytes) are added to these queues,
+  // emit the number to the UI (look for corresponding freedom.on in main.html).
+  socksRtc.bytesReceivedFromPeer.setSyncHandler((numBytes:number) => {
+    freedom.emit('newBytesReceived', numBytes);
```
trevj: naming suggestion: `bytesReceived` (and, likewise, `bytesSent`)

src/socks-server/samples/copypaste-socks-chromeapp/main.html
```diff
@@ -63,6 +63,10 @@
           that you are using uProxy, so send it via a private channel you trust.
         </p>
       </div>
+      <div>
+        <a id="giveAccessPanel_bytesSent">0</a> bytes sent to <a id="giveAccessPanel_giverName">Alice</a>.<br>
```
trevj: So, the UI doesn't otherwise reference Alice or Bob :-)

How about `x bytes received from remote party`?

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -159,6 +174,12 @@ module SocksToRtc {
 
       if(!(rtcData.channelLabel in this.sessions_)) {
         log.error('onDataFromPeer_: no such channel: ' + rtcData.channelLabel);
+        // We only log data sent via ArrayBuffer, so just check .data (and not
```
trevj: So, again, it looks like this needs to be outside of the `if` block?

(and, again, apologies if I'm wrong)

trevj:
I was a bit nit-picky but overall good work, very happy to see this in the sample app :-)

Something to think about: how could we surface this in the "Simple SOCKS" sample app? It's an important part of a more "tight-loop" development cycle.

src/rtc-to-net/rtc-to-net.ts
lucyhe: Done

src/rtc-to-net/rtc-to-net.ts
lucyhe: Done

src/rtc-to-net/rtc-to-net.ts
lucyhe: Done

src/rtc-to-net/rtc-to-net.ts
lucyhe, line 88: I create the queues where they're declared and added the comment there!

src/rtc-to-net/rtc-to-net.ts
lucyhe: Done! Thanks for clarification.

src/rtc-to-net/rtc-to-net.ts
lucyhe: Done

src/rtc-to-net/rtc-to-net.ts
lucyhe: Done

src/rtc-to-net/rtc-to-net.ts
lucyhe: Done!
(Actually since we moved the .handle to RtcToNet.onDataFromPeer_, it was removed fro this function entirely.)

src/socks-server/samples/copypaste-socks-chromeapp/freedom-module.ts
lucyhe: Done

src/socks-server/samples/copypaste-socks-chromeapp/main.html
lucyhe: Done

src/socks-to-rtc/socks-to-rtc.ts
lucyhe: Done

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -98,6 +113,8 @@ module SocksToRtc {
       }
       this.isStopped_ = true;
       this.signalsForPeer.clear();
+      this.bytesReceivedFromPeer.clear();
```
trevj, line 116: I think this probably isn't necessary, although I do see that we clear the signals queue...if you want to leave it in for whatever reason `signalsForPeer.clear()`, no problem.

trevj:
:+1:

dborkan:
w00t!!!

PR #131, opened 2014-09-25
Benchmark (with docs) for Socks5 proxy

mollyling:
image

Currently detects the data channel limit bug.


mollyling:
Sendting to trevj as a first-choice reviewer.  @trevj , please feel free to toss the potato to someone else as you feel appropriate!

Gruntfile.coffee
```diff
@@ -264,7 +288,8 @@ module.exports = (grunt) ->
     'socksCommon'
     'socksToRtc'
     'rtcToNet'
-    'socksSamples'
+    'socksSamples',
```
trevj: no comma required by coffeescript (we can simplify the diff)

Gruntfile.coffee
```diff
@@ -37,6 +37,10 @@ module.exports = (grunt) ->
         # us from being able to sym-link into node_modules and have building
         # work correctly.
         overwrite: false
+      # Fix-up a 'typings' link
```
trevj: So, I think it'd be better if we could keep the third party stuff inside `build/typescript-src/third_party/`.

I spent some time thinking while trying to make conditional compilation happen. The best/simplest I could manage was create a symlink to each direct sub-directory of `third_party/`:
https://github.com/uProxy/uproxy-lib/pull/72/files#diff-a35864fcc4749a3b904641b4be9feacfR93

Here, I guess we'd "merge" `uproxy-lib`'s and `uproxy-networking`'s `third_party/` directories.

src/benchmark/benchmark.ts
```diff
@@ -0,0 +1,384 @@
...
+        upperLimit : number;
+        count : number;
+        constructor(up : number, cnt : number) {
+            this.upperLimit = up;
+            this.count = cnt;
+        }
+    };
+
+    export class BasicStats {
+        max : number;
```
trevj: Two spaces for indentation!

src/third_party/node/node-0.8.8.d.ts
```diff
@@ -0,0 +1,1055 @@
+// Type definitions for Node.js v0.8.8
```
trevj: So, this is in tsd. I know this will be a pain while we still have separate repos (sorry) but let's use tsd for this.

In `uproxy-lib`:
- `cd third_party`
- `tsd query node --action install`

(We could also maintain a separate tsd folder in this repo but then it gets hard to merge the two tsd folders from the two repos......argh).

trevj:
Nice! I didn't get a chance to go very in-depth with the code itself so just a few comments on the build setup for now. Sorry, really wish we had just one repo too! Happy to sit down with you next week and thrash all this stuff out, then submit.

PR #133, opened 2014-09-26
Show bytes sent/received in simple socks

lucyhe:
Sample output in background page:
image

### DIFF
--- a/src/socks-server/samples/simple-socks-chromeapp/freedom-module.ts
+++ b/src/socks-server/samples/simple-socks-chromeapp/freedom-module.ts
@@ -63,25 +63,25 @@ rtcNet.signalsForPeer.setSyncHandler(socksRtc.handleSignalFromPeer);
 
 socksRtc.bytesReceivedFromPeer.setSyncHandler((numBytes:number) => {
   getterBytesReceived += numBytes;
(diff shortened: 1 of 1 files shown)


src/socks-server/samples/simple-socks-chromeapp/freedom-module.ts
```diff
@@ -52,9 +52,39 @@ var socksRtc = new SocksToRtc.SocksToRtc(
...
+var getterBytesSent :number = 0;
+var giverBytesReceived :number = 0;
+var giverBytesSent :number = 0;
+
 socksRtc.signalsForPeer.setSyncHandler(rtcNet.handleSignalFromPeer);
 rtcNet.signalsForPeer.setSyncHandler(socksRtc.handleSignalFromPeer);
 
+socksRtc.bytesReceivedFromPeer.setSyncHandler((numBytes:number) => {
+  getterBytesReceived += numBytes;
+  console.log('Getter received ' + numBytes + ' bytes. (Total received: '
```
trevj: Let's use the logger here, e.g.:

``` typescript
log.debug('Getter received ' + ... );
```

I'm suggesting debug so that the console isn't flooded by messages in the normal case. It's just good to have the option to enable these.

trevj:
Thanks! :+1: , once this is changed to use the logger.

PR #134, opened 2014-10-03
move to the uproxy-lib v12 build world

trevj:
Depends on:
https://github.com/uProxy/uproxy-churn/pull/41


trevj:
I just published the new churn NPM so this is ready to review now. Thanks!

src/ipaddrjs/ipaddrjs.d.ts
```diff
@@ -0,0 +1,73 @@
+// Type definitions for ipaddr.js 0.1.3
```
salomegeo, line 1: Why do you need to have a copy of ipaddr in this repo?

trevj, line 1: `RtcToNet` uses it to detect when a user might be trying to connect to an address on the remote user's private network, e.g. `192.168.1.1`. This TypeScript definition was in uproxy-lib but there's no real reason for it to be since it's only used here:
https://github.com/uProxy/uproxy-lib/pull/78

editor_support/sublime_text/uproxy-networking.sublime-project
```diff
@@ -1,9 +1,15 @@
 {
   "folders": [
     {
-      "path": "../../",
-      "name": "uproxy-networking project",
+      "path": "../..",
+      "name": "root",
```
iislucas: better not to use 'root' this this is the bold name that appears at the top - and root is not very distinctive. Better to use something that indicates it is uproxy-networking that way you can easily distinguish when you have multiple sublime projects open.

iislucas:
One suggestion for better text-editing experience; otherwise :+1:

editor_support/sublime_text/uproxy-networking.sublime-project
trevj: Great suggestion. Done.

PR #137, opened 2014-10-13
SOCKS server life-cycle: detailed definitions and unit tests for onceStarted and onceStopped promises

trevj:
The general idea here is that the SOCKS server depends on two things to function:
- TCP server
- peerconnection

Once both have successfully started, the SOCKS server is running (`onceStarted`); conversely, once _either_ dies, the SOCKS server has died (`onceStopped`).

This defines `onceStarted` and `onceStopped` reasonably concretely and adds unit tests for how they should behave. Those tests are obviously pretty abstract so this pull request has a number of TODOs marking where integration tests are required, e.g. verify peerconnection's `onceDisconnected` actually fires and that its `stop` method is idempodent.

Tested with the sample apps.

Clearly, a followup pull request would implement a similar lifecycle for SOCKS sessions which depend on both a TCP connection and data channel to function.

### DIFF
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -11,7 +11,6 @@ declare module SocksToRtc {
                 pcConfig?:WebRtc.PeerConnectionConfig,
                 obfuscate?:boolean);
     public stop :() => Promise<void>;
-    public onceStarted: () => Promise<void>;
     public onceReady :Promise<Net.Endpoint>;
     public isStopped :() => boolean;
     public onceStopped :() => Promise<void>;
@@ -20,11 +19,17 @@ declare module SocksToRtc {
     public bytesSentToPeer :Handler.Queue<number, void>;
     public handleSignalFromPeer :(signal: WebRtc.SignallingMessage) => void;
     public toString :() => string;
-
     public configure :(
         tcpServer:Tcp.Server,
         peerconnection:freedom_UproxyPeerConnection.Pc)
         => void;
+    public getOnceTcpServerStarted :(tcpServer:Tcp.Server) => Promise<void>;
+    public getOnceTcpServerStopped :(tcpServer:Tcp.Server) => Promise<void>;
+    public getOncePeerconnectionStarted :(
+        peerconnection:freedom_UproxyPeerConnection.Pc) => Promise<void>;
+    public getOncePeerconnectionStopped :(
+        peerconnection:freedom_UproxyPeerConnection.Pc) => Promise<void>;
+    public makeTcpToRtcSession :(tcpConnection:Tcp.Connection) => void;
   }
   class Session {
     constructor(tcpConnection:Tcp.Connection,
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -10,51 +10,30 @@
 /// <reference path="../tcp/tcp.d.ts" />
 /// <reference path="../third_party/typings/es6-promise/es6-promise.d.ts" />
 declare module SocksToRtc {
-    import WebrtcLib = freedom_UproxyPeerConnection;
     class SocksToRtc {
-        onceReady: Promise<Net.Endpoint>;
-        private onceStarted_;
         onceStarted: () => Promise<void>;
-        private onceStopped_;
+        onceReady: Promise<Net.Endpoint>;
         onceStopped: () => Promise<void>;
         signalsForPeer: Handler.Queue<WebRtc.SignallingMessage, void>;
         bytesReceivedFromPeer: Handler.Queue<number, void>;
         bytesSentToPeer: Handler.Queue<number, void>;
-        private tcpServer_;
-        private peerConnection_;
-        private sessions_;
         constructor(endpoint?: Net.Endpoint, pcConfig?: WebRtc.PeerConnectionConfig, obfuscate?: boolean);
-        setResources(tcpServer: Tcp.Server, peerconnection: WebrtcLib.Pc): void;
-        makeDefaultOnceStarted(): void;
-        makeOnceStarted(socketReady: Promise<any>, pcReady: Promise<any>): void;
-        makeDefaultOnceStopped(): void;
-        makeOnceStopped(socketStopped: Promise<any>, pcStopped: Promise<any>): void;
-        stop: () => void;
-        private cleanup_;
-        public makeTcpToRtcSession(tcpConnection:Tcp.Connection): void;
+        configure(tcpServer: Tcp.Server, peerconnection: freedom_UproxyPeerConnection.Pc): void;
+        makeOnceStarted(serverReady: Promise<any>, peerconnectionReady: Promise<any>): void;
+        makeOnceStopped(serverTerminated: Promise<any>, peerconnectionTerminated: Promise<any>): void;
+        stop: () => Promise<void>;
+        makeTcpToRtcSession: (tcpConnection: Tcp.Connection) => void;
         handleSignalFromPeer: (signal: WebRtc.SignallingMessage) => void;
-        private onDataFromPeer_;
         toString: () => string;
     }
     class Session {
-        tcpConnection: Tcp.Connection;
-        private peerConnection_;
-        private bytesReceivedFromPeer;
-        private bytesSentToPeer;
-        private channelLabel_;
         onceReady: Promise<Net.Endpoint>;
         onceClosed: Promise<void>;
-        private dataChannelIsClosed_;
-        private dataFromPeer_;
-        constructor(tcpConnection: Tcp.Connection, peerConnection_: WebrtcLib.Pc, bytesReceivedFromPeer: Handler.Queue<number, void>, bytesSentToPeer: Handler.Queue<number, void>);
+        constructor(tcpConnection: Tcp.Connection, peerConnection_: freedom_UproxyPeerConnection.Pc, bytesReceivedFromPeer: Handler.Queue<number, void>, bytesSentToPeer: Handler.Queue<number, void>);
         longId: () => string;
         close: () => Promise<void>;
         handleDataFromPeer: (data: WebRtc.Data) => void;
         channelLabel: () => string;
         toString: () => string;
-        private doAuthHandshake_;
-        private receiveEndpointFromPeer_;
-        private doRequestHandshake_;
-        private linkTcpAndPeerConnectionData_;
     }
 }
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -1,39 +1,36 @@
-/// <reference path="../socks-common/socks-headers.d.ts" />
-/// <reference path="../freedom/coreproviders/uproxylogging.d.ts" />
 /// <reference path="../freedom/coreproviders/uproxypeerconnection.d.ts" />
-/// <reference path="../freedom/typings/freedom.d.ts" />
 /// <reference path="../handler/queue.d.ts" />
 /// <reference path="../networking-typings/communications.d.ts" />
-/// <reference path="../churn/churn.d.ts" />
 /// <reference path="../webrtc/datachannel.d.ts" />
 /// <reference path="../webrtc/peerconnection.d.ts" />
 /// <reference path="../tcp/tcp.d.ts" />
 /// <reference path="../third_party/typings/es6-promise/es6-promise.d.ts" />
 declare module SocksToRtc {
-    class SocksToRtc {
-        onceStarted: () => Promise<void>;
-        onceReady: Promise<Net.Endpoint>;
-        onceStopped: () => Promise<void>;
-        signalsForPeer: Handler.Queue<WebRtc.SignallingMessage, void>;
-        bytesReceivedFromPeer: Handler.Queue<number, void>;
-        bytesSentToPeer: Handler.Queue<number, void>;
-        constructor(endpoint?: Net.Endpoint, pcConfig?: WebRtc.PeerConnectionConfig, obfuscate?: boolean);
-        configure(tcpServer: Tcp.Server, peerconnection: freedom_UproxyPeerConnection.Pc): void;
-        makeOnceStarted(serverReady: Promise<any>, peerconnectionReady: Promise<any>): void;
-        makeOnceStopped(serverTerminated: Promise<any>, peerconnectionTerminated: Promise<any>): void;
-        stop: () => Promise<void>;
-        makeTcpToRtcSession: (tcpConnection: Tcp.Connection) => void;
-        handleSignalFromPeer: (signal: WebRtc.SignallingMessage) => void;
-        toString: () => string;
-    }
-    class Session {
-        onceReady: Promise<Net.Endpoint>;
-        onceClosed: Promise<void>;
-        constructor(tcpConnection: Tcp.Connection, peerConnection_: freedom_UproxyPeerConnection.Pc, bytesReceivedFromPeer: Handler.Queue<number, void>, bytesSentToPeer: Handler.Queue<number, void>);
-        longId: () => string;
-        close: () => Promise<void>;
-        handleDataFromPeer: (data: WebRtc.Data) => void;
-        channelLabel: () => string;
-        toString: () => string;
-    }
+  class SocksToRtc {
+    constructor(endpoint:Net.Endpoint,
+                pcConfig:WebRtc.PeerConnectionConfig,
+                obfuscate:boolean);
+    public stop :() => Promise<void>;
+    public onceStarted: () => Promise<void>;
+    public onceReady :Promise<Net.Endpoint>;
+    public isStopped :() => boolean;
+    public onceStopped :() => Promise<void>;
+    public signalsForPeer :Handler.Queue<WebRtc.SignallingMessage, void>;
+    public bytesReceivedFromPeer :Handler.Queue<number, void>;
+    public bytesSentToPeer :Handler.Queue<number, void>;
+    public handleSignalFromPeer :(signal: WebRtc.SignallingMessage) => void;
+    public toString :() => string;
+  }
+  class Session {
+    constructor(tcpConnection:Tcp.Connection,
+                peerConnection_:freedom_UproxyPeerConnection.Pc);
+    public tcpConnection :Tcp.Connection;
+    public onceReady :Promise<Net.Endpoint>;
+    public onceClosed :Promise<void>;
+    public longId :() => string;
+    public close :() => Promise<void>;
+    public handleDataFromPeer :(data:WebRtc.Data) => void;
+    public channelLabel :() => string;
+    public toString :() => string;
+  }
 }
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -23,12 +23,6 @@ declare module SocksToRtc {
         tcpServer:Tcp.Server,
         peerconnection:freedom_UproxyPeerConnection.Pc)
         => Promise<Net.Endpoint>;
-    public getOnceTcpServerStarted :(tcpServer:Tcp.Server) => Promise<void>;
-    public getOnceTcpServerStopped :(tcpServer:Tcp.Server) => Promise<void>;
-    public getOncePeerconnectionStarted :(
-        peerconnection:freedom_UproxyPeerConnection.Pc) => Promise<void>;
-    public getOncePeerconnectionStopped :(
-        peerconnection:freedom_UproxyPeerConnection.Pc) => Promise<void>;
     public makeTcpToRtcSession :(tcpConnection:Tcp.Connection) => void;
   }
   class Session {
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -20,18 +20,10 @@ declare module SocksToRtc {
     public bytesSentToPeer :Handler.Queue<number, void>;
     public handleSignalFromPeer :(signal: WebRtc.SignallingMessage) => void;
     public toString :() => string;
-    public setResources :(
+
+    public configure :(
         tcpServer:Tcp.Server,
         peerconnection:freedom_UproxyPeerConnection.Pc)
-        => void;
-    public configure :() => void;
-    public makeOnceStarted :(
-        serverReady:Promise<any>,
-        peerconnectionReady:Promise<any>)
-        => void;
-    public makeOnceStopped :(
-        serverTerminated:Promise<any>,
-        peerconnectionTerminated:Promise<any>)
         => void;
   }
   class Session {
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -7,9 +7,9 @@
 /// <reference path="../third_party/typings/es6-promise/es6-promise.d.ts" />
 declare module SocksToRtc {
   class SocksToRtc {
-    constructor(endpoint:Net.Endpoint,
-                pcConfig:WebRtc.PeerConnectionConfig,
-                obfuscate:boolean);
+    constructor(endpoint?:Net.Endpoint,
+                pcConfig?:WebRtc.PeerConnectionConfig,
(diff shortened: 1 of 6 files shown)


src/socks-to-rtc/socks-to-rtc.d.ts
```diff
@@ -1,35 +1,39 @@
+/// <reference path="../socks-common/socks-headers.d.ts" />
+/// <reference path="../freedom/coreproviders/uproxylogging.d.ts" />
 /// <reference path="../freedom/coreproviders/uproxypeerconnection.d.ts" />
+/// <reference path="../freedom/typings/freedom.d.ts" />
 /// <reference path="../handler/queue.d.ts" />
 /// <reference path="../networking-typings/communications.d.ts" />
+/// <reference path="../churn/churn.d.ts" />
```
iislucas: Churn isn't needed or used as far as I can see. Similar for freedom and socks?

src/socks-to-rtc/socks-to-rtc.d.ts
```diff
@@ -1,35 +1,39 @@
...
+    class SocksToRtc {
+        onceStarted: () => Promise<void>;
+        onceReady: Promise<Net.Endpoint>;
+        onceStopped: () => Promise<void>;
+        signalsForPeer: Handler.Queue<WebRtc.SignallingMessage, void>;
+        bytesReceivedFromPeer: Handler.Queue<number, void>;
+        bytesSentToPeer: Handler.Queue<number, void>;
+        constructor(endpoint?: Net.Endpoint, pcConfig?: WebRtc.PeerConnectionConfig, obfuscate?: boolean);
+        configure(tcpServer: Tcp.Server, peerconnection: freedom_UproxyPeerConnection.Pc): void;
+        makeOnceStarted(serverReady: Promise<any>, peerconnectionReady: Promise<any>): void;
```
iislucas: I don't think you want these to be any types.
same for below. Maybe void?

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+      .then(done);
+  });
+
+  it('onceStarted rejects and cleans up on socket setup failure', (done) => {
+    server.makeOnceStarted(
+        Promise.reject(new Error('failed to listen')), // socket
+        new Promise((F, R) => {}));                    // peerconnection
+    server.onceStarted()
+      .catch((e:Error) => {
+        expect(stop).toHaveBeenCalled();
```
iislucas: Do we really want the spec to force stop to be called, or do we want to check that the stop pomise is fulfilled? If possible, would be ideal for test to be tied to the public data available rather than to internal function call paths... unless we want to make explicit (in comments) in the .d.ts file that failure results in a call of stop...?

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+        Promise.reject(new Error('failed to negotiate'))); // peerconnection
+    server.onceStarted()
+      .catch((e:Error) => {
+        expect(stop).toHaveBeenCalled();
+      })
+      .then(done);
+  });
+
+  it('onceStopped fulfills and cleans up on socket termination fulfillment', (done) => {
+    server.makeOnceStopped(
```
iislucas: I'm a little surprised that onceStopped succeeds independently of onceStarted... maybe worth some discussion.

src/socks-to-rtc/socks-to-rtc.d.ts
```diff
@@ -1,35 +1,39 @@
...
+        makeOnceStopped(serverTerminated: Promise<any>, peerconnectionTerminated: Promise<any>): void;
```
iislucas: I think we need some comments here to explain what makeOnceStopped does/why it exists. Likewise for makeOnceStarted; and especially to explain how they are related to onceSTarted and onceStopped.

src/socks-to-rtc/socks-to-rtc.d.ts
```diff
@@ -1,35 +1,39 @@
...
-    public onceClosed :Promise<void>;
-    public longId :() => string;
-    public close :() => Promise<void>;
-    public handleDataFromPeer :(data:WebRtc.Data) => void;
-    public channelLabel :() => string;
-    public toString :() => string;
-  }
+    class SocksToRtc {
+        onceStarted: () => Promise<void>;
+        onceReady: Promise<Net.Endpoint>;
```
iislucas: While we're doing some cleanup, we should make onceReady into a function of `type () => Promise<Net.Endpoint>;` better for testing, chaining promises, and avoids external setting.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -28,16 +26,32 @@ module SocksToRtc {
...
   export class SocksToRtc {
-    // Holds the IP/port that the localhost socks server is listeneing to.
-    public onceReady : Promise<Net.Endpoint>;
 
-    private isStopped_ :boolean;
-    public isStopped = () : boolean => { return this.isStopped_; }
+    // Fulfills once the TCP socket is listening for connections and a
+    // peerconnection has been successfully established. Rejects if either
+    // socket or peerconnection setup fails. On rejection, both TCP socket
+    // and peerconnection will be closed.
```
iislucas: The unit test above isn't quite testing this spec; it tests that the global stop method is called. But I think this comment is right. Maybe good to discuss how to get the full tests we want. I suspect that correctly mocking out used classes is the way to go.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -28,16 +26,32 @@ module SocksToRtc {
...
+    private onceStarted_ :Promise<void>;
+    public onceStarted = () : Promise<void> => { return this.onceStarted_; }
+
+    // As onceStarted(), but reports the address on which the server
+    // is listening for connections.
+    // TODO: Remove this public field in favour of onceStarted() and a
```
iislucas: Not sure... the getEndpoint would have to return a promise, and then that would basically be the same as onceReady... I suspect the only thing needed is a single onceStarted_/onceReady promise.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -28,16 +26,32 @@ module SocksToRtc {
...
+    //       getEndpoint()-type method
+    public onceReady :Promise<Net.Endpoint>;
+
+    // Fulfills once the TCP socket has closed or the peerconnection
```
iislucas: great comment.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -28,16 +26,32 @@ module SocksToRtc {
...
+    // has terminated *and* there is subsequently no error closing the
+    // socket or the peerconnection (if an error is encountered at this
+    // stage, this rejects).
+    // Behaviour is undefined if onceStarted() has rejected so do not
```
iislucas: Better to define it no? onceStopped should reject is startup fails would be a natural expectation and decrease undefined inter-dependencies between variables. (better to have things always defined if possible)

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -65,92 +85,127 @@ module SocksToRtc {
...
-      // the tcpo server too, and do any needed cleanup.
-      this.onceStopped_ = this.peerConnection_.onceDisconnected()
-          .then(this.stop);
-
-      // Return promise for then we have the tcp-server endpoint & we have a
-      // peer connection.
-      this.onceReady = oncePeerConnectionReady
-        .then(() => { return onceTcpServerReady; });
+      if (endpoint) {
+        this.configure(
```
iislucas: This seems to be the only code-path to calling configure... I suspect there is a testing related reason for having configure and making constructor arguments optional... ? :)

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -65,92 +85,127 @@ module SocksToRtc {
...
-    // connection. TODO: If the given port is zero, platform chooses a port and
-    // this listening port is returned by the |onceReady| promise.
+    private sessions_ :{ [channelLabel:string] : Session } = {};
+
+    // Creates a new SOCKS server running on the specified address.
+    // A TCP server and peerconnection, configured with the supplied endpoint
+    // and config, are constructed.
+    // If the endpoint is undefined, the caller must manually configure the
+    // server; this is only intended for unit tests.
+    // TODO: Replace this with a static constructor.
```
iislucas: why is that better?

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -65,92 +85,127 @@ module SocksToRtc {
...
+        }, (e:Error) => {
+          return Promise.reject(e);
         });
     }
 
-    // Setup a SOCKS5 TCP-to-rtc session from a tcp connection.
-    private makeTcpToRtcSession_ = (tcpConnection:Tcp.Connection) : void => {
+    // Invoked when a SOCKS client establishes a connection with our
+    // server socket.
+    public makeTcpToRtcSession = (tcpConnection:Tcp.Connection) : void => {
```
iislucas, line 159: why public? Doesn't seem anyone else calls it?

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -65,92 +85,127 @@ module SocksToRtc {
...
+          this.peerConnection_.negotiateConnection());
+
+      // TODO: Integration tests for these objects' termination behaviours.
+      this.makeOnceStopped(
+          // TCP server has no onceDisconnected()-type method!
+          // So, supply a dummy promise that neither resolves nor rejects.
+          new Promise((F, R) => {}),
+          this.peerConnection_.onceDisconnected());
+
+      this.onceReady = this.onceStarted_.then(() => {
```
iislucas: best to merge onceReady and onceStarted I think. I don't see advantage in their separation.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -65,92 +85,127 @@ module SocksToRtc {
...
+        return {
+          address: this.tcpServer_.endpoint.address,
+          port: this.tcpServer_.endpoint.port
+        };
       });
-      this.peerConnection_.on('signalForPeer',
-          this.signalsForPeer.handle);
-
-      var onceConnected = this.peerConnection_.onceConnected()
```
iislucas, line 139: You need this I'm afraid. Otherwise initial peer connection is created without data channels, and then you get a second round of signalling messages. Chat2 demo showed this bad behaviour. We may want it back to catch this kind of thing. I forgot about that.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -65,92 +85,127 @@ module SocksToRtc {
...
+    // Public for unit tests, so that termination failures can be easily
+    // simulated.
+    public makeOnceStopped(
+        serverTerminated:Promise<any>,
+        peerconnectionTerminated:Promise<any>)
+        : void {
+      if (this.onceStopped_) {
+        throw new Error('onceStopped_ already set');
+      }
+      this.onceStopped_ = Promise.race([
```
iislucas: Not clear to me you want a race here. e.g. if you race, and PC closed before TCP socket and then you create a new TCP socket on old port, you'll confusingly get a race condition. Better to know you've stopped when both are stopped.

iislucas:
Awesome to have these tests, improved comments, and some good hard thinking on the flow structure. We should do a longer sit-down discussion as there's a few tricky things here. But I like where it's going.

src/socks-to-rtc/socks-to-rtc.d.ts
trevj: So, I've reverted this file except to add the new public methods I've added to the implementation.

src/socks-to-rtc/socks-to-rtc.d.ts
trevj: My intention here with `any` is to say "I don't care with what value your promise fulfills, so long as it fulfills once the server is ready". In other words, a Promise of any type.

src/socks-to-rtc/socks-to-rtc.d.ts
trevj: These are commented in the implementation.

src/socks-to-rtc/socks-to-rtc.spec.ts
trevj: Great question. I was trying to mock as little as possible, and just checking for a call to `stop` lets me mock just a tiny bit of the TCP server and peerconnection. However, I should also have tests for `stop` itself.

So, I've added tests for `stop` and made it explicit in `onceStarted`'s comments that it calls `stop` on rejection. How does that sound?

src/socks-to-rtc/socks-to-rtc.ts
trevj: So, I've changed this comment to state that `stop()` will be called in the event of rejection.

src/socks-to-rtc/socks-to-rtc.ts
trevj: Yes! This is explained in the comments in the implementation.

src/socks-to-rtc/socks-to-rtc.ts
trevj: Two reasons it's preferable to separate the two concerns:
- simplifies tests (why should `onceStarted` tests have to worry about endpoints)
- endpoint may not be the most important return value in the future

I think the better question is why merge the two? :-)

src/socks-to-rtc/socks-to-rtc.ts
trevj: :-)

src/socks-to-rtc/socks-to-rtc.ts
trevj: Here I was thinking:
1. static constructors can return objects
2. facilitates overloading, for unit tests

src/socks-to-rtc/socks-to-rtc.ts
trevj: See my comment elsewhere on this.

src/socks-to-rtc/socks-to-rtc.ts
trevj, line 159: Ah, any caller who wants to avoid the constructor and configure the object manually would need access to this method.

src/socks-to-rtc/socks-to-rtc.ts
trevj: So, I think this is what we want.

`race` will call `stop` when _either_ TCP server or peerconnection terminates. `stop` will then fulfill (or reject) once _both_ TCP server and peerconnection have terminated.

What sort of race condition were you envisioning -- how would I create a new TCP socket on the old port?

src/socks-to-rtc/socks-to-rtc.ts
trevj, line 139: So, which part exactly do we need? I don't see the extra round of messages you're talking about (on Chrome 39: one SDP + one candidate sent from each side to the other).

trevj:
Many thanks for the detailed review!

I think your biggest outstanding question is whether `onceStopped` should fulfill if `onceStarted` fulfills. I had that thought too, initially, but when I attempted to code it only complicated my tests without providing a clear benefit. Let me think about that a little more.

src/socks-to-rtc/socks-to-rtc.spec.ts
iislucas: If we are testing that `stop` is called, we have to be making the call that that is part of the specification (We definitely shouldn't be minimizing our mocks at the expense of testing things that we don't want to be true!). Generally, that's not good style for a specification to test behaviour instead of state. How about adding a function call to actually get the connection state, e.g. `isReady() : boolean`. Then we check the value of isReady?

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  });
+
+  it('onceStarted rejects and cleans up on peerconnection setup failure', (done) => {
+    var stop = spyOn(server, 'stop');
+    server.makeOnceStarted(
+        new Promise((F, R) => {}),                         // socket
+        Promise.reject(new Error('failed to negotiate'))); // peerconnection
+    server.onceStarted()
+      .catch((e:Error) => {
+        expect(stop).toHaveBeenCalled();
```
iislucas: Whatever resolution we get to above should be fixed here too.

src/socks-to-rtc/socks-to-rtc.ts
iislucas: I don't see why tests are simpler; you can simply ignore the returned promise value. 

What other value might we want in the future? Wouldn't it be better to return whatever information is important about starting up when we know it/add it? 

The reason in my mind to merge them is that `started` and `ready` sound like the same thing, and in fact internally are the same thing, one is just ignoring a value of the other.

src/socks-to-rtc/socks-to-rtc.ts
iislucas: The alternative is that onceStopped should neither be rejected nor fulfilled. But that easy to lead to hanging calls and memory leaks. Better to have defined failure.

src/socks-to-rtc/socks-to-rtc.ts
iislucas: I see, the idea being to make constructors as simple as possible and not actually having any logic you can't mock-out easily?

I'm not sure about this design pattern... I'd be inclined to have proper mocks for all objects instead... You produce more readable code using constructors to do the constructing... tricky one!

src/socks-to-rtc/socks-to-rtc.ts
iislucas: OK, let's resolve it here once we decide above :)

src/socks-to-rtc/socks-to-rtc.ts
iislucas, line 139: When you later open a data channel, it will then make you send another SDP header. I complained about this to WebRTC, but seems to be part of the spec.

Also, I just did some experiments and as it happens, it seems we may need a control data channel in order to know if/when the peer-connection dies (spec needs investigation; see: https://github.com/uProxy/uproxy/issues/408 ).

src/socks-to-rtc/socks-to-rtc.ts
iislucas: Oh, yes, you're right, I was mis-reading. 

So the disadvantage of this style is that it ends up calling close on tcp and peerconnection twice. There was a little work to avoid that before. This is basically a life-cycle design pattern question. Last time I reviewed it, it was better to avoid double closures / know when you do them that you have a problem.  I'm guessing you felt like there was too much work there. I'd suggest we think about abstracting the mutual-close functionality into a generic class actually. We want to do the same thing in rtc-to-net also. 

Probably needs more discussion / a whiteboard session?

src/socks-to-rtc/socks-to-rtc.ts
iislucas, line 159: I see... this is part of the idea of being able to more easily make stubs of construction, but we just havn't got anyone else calling it yet?

trevj, line 159: That's right.

trevj:
OK, I re-worked this quite a bit.

Let's forget about the `onceReady`/`onceStarted` business for now.

I refactored the code so I can easily mock the functions which will require integration testing: they create promises which fulfill when the peerconnection starts, stops, etc.

`onceReady` is the same as before.

For `onceStopped`, I found it useful to introduce a `onceStopping_` promise. This lets us initiate shutdown in a variety of ways, and helps us call shutdown on the TCP server and peerconnection exactly once. Shutdown can be initiated for three reasons:
- startup failure
- TCP server or peerconnection failure
- manual invocation of stop()

If you don't strongly object to any of this, I will start working on a similar approach for SOCKS sessions.

src/socks-to-rtc/socks-to-rtc.d.ts
```diff
@@ -7,10 +7,11 @@
...
 declare module SocksToRtc {
   class SocksToRtc {
-    constructor(endpoint:Net.Endpoint,
-                pcConfig:WebRtc.PeerConnectionConfig,
-                obfuscate:boolean);
+    constructor(endpoint?:Net.Endpoint,
+                pcConfig?:WebRtc.PeerConnectionConfig,
+                obfuscate?:boolean);
     public stop :() => Promise<void>;
+    public onceStarted: () => Promise<void>;
```
iislucas: You probably want to remove this now there is no implementation.

src/socks-to-rtc/socks-to-rtc.d.ts
```diff
@@ -19,6 +20,11 @@ declare module SocksToRtc {
     public bytesSentToPeer :Handler.Queue<number, void>;
     public handleSignalFromPeer :(signal: WebRtc.SignallingMessage) => void;
     public toString :() => string;
+
+    public configure :(
```
iislucas: The configure function actually starts the server listening; what do you think about calling it 'start' and returning the onceStarted promise?

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -65,92 +77,122 @@ module SocksToRtc {
...
-      return onceConnected.then(() => {
-          return this.peerConnection_.openDataChannel('_control_')
-        })
-        .then(() => {
-          this.peerConnection_.send('_control_', { str: 'hello?' });
-          return onceConnected;
+      this.peerConnection_.on('signalForPeer', this.signalsForPeer.handle);
+
+      // Start the peerconnection. TCP server doesn't have a start/onceReady
+      // separation so we let getOnceTcpServerStarted start the TCP server.
```
iislucas: Not sure what the comment is talking about TCP Server for...?

src/socks-to-rtc/socks-to-rtc.d.ts
trevj: Good point. I added all the new public functions here.

src/socks-to-rtc/socks-to-rtc.d.ts
trevj: I think I like this. I've done that and done the same for `stop` (returns `onceStopped_`).

src/socks-to-rtc/socks-to-rtc.ts
trevj: Just me being overly verbose. I shortened this.

trevj:
So I took your suggestion of removing those `getOnceXXXStarted()` methods in favour of mocking the relevant `start()` and `stop()` methods on the TCP server and peerconnection objects.

I like this a lot better now: the tests are shorter and the interface for SOCKS server a lot nicer.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -65,92 +83,79 @@ module SocksToRtc {
...
-          return this.peerConnection_.openDataChannel('_control_')
-        })
-        .then(() => {
-          this.peerConnection_.send('_control_', { str: 'hello?' });
-          return onceConnected;
+      this.peerConnection_.on('signalForPeer', this.signalsForPeer.handle);
+
+      // Start and listen for notifications.
+      peerconnection.negotiateConnection();
+      this.onceReady = Promise.all<any>([
```
iislucas, line 120: I think I prefer something like: 

```
tcpServer.listen();
this.onceReady = peerconnection.onceConnected().then(tcpServer.onceListening);
```

(The array-binding to first element via an any-type coercion trick to get the result of the tcpServer listen is tricky: it's easy to get that wrong and the compiler won't help you)

I've added `tcpServer.onceListening` so you can do this... see this branch:  https://github.com/uProxy/uproxy-networking/blob/iislucas-tcp/src/tcp/tcp.d.ts

iislucas:
:+1:  Looking great! 

One small idea for slight code niceity, but depends on an unmerged pull-request, so maybe push that cleanup to the forthcoming iislucas-tcp pull request....?

src/socks-to-rtc/socks-to-rtc.ts
trevj, line 120: I've added a TODO to use `tcpServer.onceListening` once it's available. I would prefer that too.

Regarding `Promise<any>`, what you suggest isn't _exactly_ the same as `Promise.all` because it won't fail fast. What I can do is ignore the returned array and instead return `tcpServer.endpoint`.

PR #139, opened 2014-10-20
Localize strings in copypaste SOCKS sample

lucyhe:
ran grunt test

There is a drop down that switches the language between Italian and English.
After migrating the popup to Polymer I can try a RTL language.


src/samples/copypaste-socks-chromeapp/locales/en/messages.json
```diff
@@ -0,0 +1,53 @@
+{
```
dborkan, line 1: If the i18n API doesn't care, can we change the tabs to 2 space characters?

src/samples/copypaste-socks-chromeapp/locales/en/messages.json
```diff
@@ -0,0 +1,53 @@
...
+	"introMsg" : {
```
dborkan: Is the Msg suffix neccessary?  Seems like we could remove it to make things a little more readable

src/samples/copypaste-socks-chromeapp/main.ts
```diff
@@ -174,3 +175,39 @@ freedom.on('bytesSent', (numNewBytesSent:number) => {
...
+        translations[key] = translations[key].message;
+      }
+    }
+    i18nTemplate.process(document, translations);
+  }
+  xhr.send(null);  
+}
+
+// Listen for events indicating the language has changed.
+getLanguageInputNode.onchange = function(event:Event) : any {
```
dborkan: you can change the return type here from : any to : void

src/samples/copypaste-socks-chromeapp/locales/en/messages.json
lucyhe, line 1: done

src/samples/copypaste-socks-chromeapp/locales/en/messages.json
lucyhe: done

src/samples/copypaste-socks-chromeapp/main.ts
lucyhe: done

third_party/lib/i18n/i18n.d.ts
```diff
@@ -0,0 +1,5 @@
+interface I18n {
```
trevj: I don't think there's any need to have `third_party/lib/i18n` -- let's just push it up to `third_party/i18n`.

third_party/lib/i18n/i18n_template.js
```diff
@@ -0,0 +1,121 @@
+// Copyright (c) 2010 The Chromium Authors. All rights reserved.
```
trevj: I think it would be good to add a README documenting where this files comes from so that we can (try to) keep it updated in future.

src/samples/copypaste-socks-chromeapp/main.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference path='../../freedom/typings/freedom.d.ts' />
 /// <reference path='../../webrtc/peerconnection.d.ts' />
+/// <reference path='lib/i18n/i18n.d.ts' />
```
trevj: Is this right? I would have expected:

``` typescript
/// <reference path='../../third_party/i18n/i18n.d.ts' />
```

Gruntfile.coffee
```diff
@@ -45,6 +45,15 @@ module.exports = (grunt) ->
           filter: 'isDirectory'
           dest: 'build/third_party/'
         ]
+      
+      # Symlink i18n from the third_party directory
```
trevj: So, rather than a separate rule for this one folder let's add a `symlink:thirdParty` rule to this Gruntfile, just like uproxy-lib's:
https://github.com/uProxy/uproxy-lib/blob/dev/Gruntfile.coffee#L96

Bug me if this doesn't work.

third_party/lib/i18n/i18n.d.ts
trevj: I suspect before too long it would be good to have this in uproxy-lib. No pressing need to do it before you're happier with the whole localisation process, though.

trevj:
So I'm mostly just reviewing the build process stuff. I have one or two comments on how we can improve that, then we should be good.

third_party/lib/i18n/i18n.d.ts
lucyhe: done (moved out of /lib)

third_party/lib/i18n/i18n_template.js
lucyhe: done

src/samples/copypaste-socks-chromeapp/main.ts
lucyhe: done

Gruntfile.coffee
lucyhe: done

third_party/lib/i18n/i18n_template.js
trevj: :+1:

src/samples/copypaste-socks-chromeapp/main.ts
trevj: :+1:

Gruntfile.coffee
```diff
@@ -46,6 +46,16 @@ module.exports = (grunt) ->
           dest: 'build/third_party/'
         ]
 
+      # Symlink each directory under third_party/ under build/third_party/.
+      thirdParty:
```
trevj: super-nit: move this to just after `symlink:build`

Gruntfile.coffee
```diff
@@ -226,6 +237,7 @@ module.exports = (grunt) ->
   taskManager.add 'copypasteSocksChromeApp', [
     'base'
     'socks'
+    'symlink:thirdParty'
```
trevj: Let's make this part of `base`. Sooner or later some other piece of `third_party` code will appear here and I'd like to keep this as similar to `uproxy-lib` as possible to facilitate possible future merging of the repos.

Gruntfile.coffee
trevj: A couple of teeny tiny nits and then :+1: . Good work!

Gruntfile.coffee
lucyhe: done

Gruntfile.coffee
lucyhe: done

PR #140, opened 2014-10-21
improving/filling out the tcp interface, proper closing

iislucas:
- Added echo-server stresser bash script to the repository
- Moved to using a handler queue for connections to the tcp server
  Instead of 

```
    tcpServer_ = new Tcp.Server(endpoint, function);
```

  You now write: 

```
    tcpServer_ = new Tcp.Server(endpoint);
      tcpServer_.connectionsQueue.setSyncHandler(
          makeTcpToRtcSession_);
```

  And of course you can do whatever you want with the handler queue, as usual. 
- Added `onceListening` and `onceShutdown` promises. 
- Added isListening an isShutdown synchronous checking functions. 
- Improved comments
- Fixed closing of freedom's tcpsocket interfaces
- Commented out logging statements (which cause bad slowdown with so many)

Fixed: https://github.com/uProxy/uproxy/issues/489
Fixed: https://github.com/uProxy/uproxy/issues/485 

TESTED: 
grunt test
Manually ran the `socks-chrome-app`

### DIFF
--- a/src/tcp/tcp.d.ts
+++ b/src/tcp/tcp.d.ts
@@ -8,29 +8,47 @@
 declare module Tcp {
   class Server {
     constructor(endpoint        :Net.Endpoint,
-                onConnection    :(c:Connection) => void,
                 maxConnections  ?:number);
 
-    public  endpoint        :Net.Endpoint;
+    // The max allowed number of connections. More than this many connections
+    // will result in new connections being closed as soon as they connect.
+    public maxConnections  :number;
 
-    public connections :() => Connection[];
-    public connectionsCount :() => number;  // equivalent to connections().length
-    public closeAll :() => Promise<void>;  // calls close on all connections.
+    // The endpoint this server is listening on.
+    public connectionsQueue :Handler.Queue<Connection, void>;
 
-    // start accepting new connections; returns the actual endpoint it ends up
+    // Gets all the current connections to this server.
+    public currentConnections :() => Connection[];
+    // Equivalent to connections().length
+    public connectionsCount :() => number;
+
+    // Start accepting new connections; returns the actual endpoint it ends up
     // listening on (e.g. when port 0 is specifed, a dynamic port number is
     // chosen).
     public listen :() => Promise<Net.Endpoint>;
-    // stop accepting new connections.
+    // Getter for the once listening promise. Use to write prettier code.
+    public onceListening :() => Promise<Net.Endpoint>;
+    // The |isListening| variable is
+    public isListening :() => boolean;
+    // Stop accepting new connections.
     public stopListening :() => Promise<void>;
 
-    public shutdown :() => Promise<void>; // stop listening and then close-all
+    // Calls close on all connections. Doesn't stop listening.
+    public closeAll :() => Promise<void>;
+
+    // Stop listening and then close-all
+    public shutdown :() => Promise<void>;
+    // The |onceShutdown| promise can be fulfilled by either a call to
+    // shutdown, or by something going wrong in the OS.
+    public onceShutdown :() => Promise<void>
+    // Synchronous access to is |onceShutdown| fulfilled.
+    public isShutdown :() => boolean
 
     // Mostly useful for debugging
     public toString :() => string;
   }
 
-  // Code for how a Tcp Connection is closed.
+  // Describes how a Tcp Connection got to the closed state.
   enum SocketCloseKind {
     WE_CLOSED_IT,
     REMOTELY_CLOSED,
@@ -38,11 +56,7 @@ declare module Tcp {
     UNKOWN
   }
 
-  /**
-  * Tcp.Connection - Wraps up a single TCP connection to a client
-  *
-  * @param {number} socketId The ID of the server<->client socket.
-  */
+  // Wraps up a single TCP connection to a client
   class Connection {
     constructor(connectionKind :Connection.Kind);
 
@@ -69,6 +83,7 @@ declare module Tcp {
     public connectionId: string;
     public getState :() => Connection.State;
     public isClosed :() => boolean;
+
     public toString :() => string;
   }
 
--- a/src/tcp/tcp.d.ts
+++ b/src/tcp/tcp.d.ts
@@ -14,7 +14,8 @@ declare module Tcp {
     // will result in new connections being closed as soon as they connect.
     public maxConnections  :number;
 
-    // The endpoint this server is listening on.
+    // The handler queue of connections made to this server. By default no
+    // handler is set; the consumer of the server is expected to set one.
     public connectionsQueue :Handler.Queue<Connection, void>;
 
     // Gets all the current connections to this server.
@@ -24,11 +25,12 @@ declare module Tcp {
 
     // Start accepting new connections; returns the actual endpoint it ends up
     // listening on (e.g. when port 0 is specifed, a dynamic port number is
-    // chosen).
+    // chosen). Returns the same promise as calling `onceListening`.
     public listen :() => Promise<Net.Endpoint>;
     // Getter for the once listening promise. Use to write prettier code.
     public onceListening :() => Promise<Net.Endpoint>;
-    // The |isListening| variable is
+    // The |isListening| variable is true after |onceListening| and before
+    // |onceShutdown|
     public isListening :() => boolean;
     // Stop accepting new connections.
     public stopListening :() => Promise<void>;
@@ -36,7 +38,8 @@ declare module Tcp {
     // Calls close on all connections. Doesn't stop listening.
     public closeAll :() => Promise<void>;
 
-    // Stop listening and then close-all
+    // Stops accepting new connection (like |stopListening|) and then closes
+    // all connections (like |closeAll|).
     public shutdown :() => Promise<void>;
     // The |onceShutdown| promise can be fulfilled by either a call to
     // shutdown, or by something going wrong in the OS.
--- a/src/tcp/tcp.d.ts
+++ b/src/tcp/tcp.d.ts
@@ -28,6 +28,8 @@ declare module Tcp {
     // chosen). Returns the same promise as calling `onceListening`.
     public listen :() => Promise<Net.Endpoint>;
     // Getter for the once listening promise. Use to write prettier code.
+    // Fulfills once the server has server has successfully bound to a port
+    // and is accepting connections. Rejects if there is any error.
     public onceListening :() => Promise<Net.Endpoint>;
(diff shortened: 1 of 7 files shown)


iislucas:
@trevj 

This is slightly more than you had in mind... but I think it's good... of course next up is lots of tests. Want those in this, or separate?

also the fix to the socks all closing based on it...?

src/tcp/tcp.d.ts
```diff
@@ -8,29 +8,47 @@
...
 
-    public  endpoint        :Net.Endpoint;
+    // The max allowed number of connections. More than this many connections
+    // will result in new connections being closed as soon as they connect.
+    public maxConnections  :number;
 
-    public connections :() => Connection[];
-    public connectionsCount :() => number;  // equivalent to connections().length
-    public closeAll :() => Promise<void>;  // calls close on all connections.
+    // The endpoint this server is listening on.
```
trevj: This is not the comment you're looking for :-)

src/tcp/tcp.d.ts
```diff
@@ -8,29 +8,47 @@
...
+    public currentConnections :() => Connection[];
+    // Equivalent to connections().length
+    public connectionsCount :() => number;
+
+    // Start accepting new connections; returns the actual endpoint it ends up
     // listening on (e.g. when port 0 is specifed, a dynamic port number is
     // chosen).
     public listen :() => Promise<Net.Endpoint>;
-    // stop accepting new connections.
+    // Getter for the once listening promise. Use to write prettier code.
```
trevj, line 30:
``` js
// Fulfills once the server has server has successfully bound to a port
// and is accepting connections. Rejects if there is any error.
```

And mention that `listen()` returns this.

src/tcp/tcp.d.ts
```diff
@@ -8,29 +8,47 @@
...
+    public onceListening :() => Promise<Net.Endpoint>;
+    // The |isListening| variable is
```
trevj: Comment needs fixing.

src/tcp/tcp.d.ts
```diff
@@ -8,29 +8,47 @@
...
+    public isListening :() => boolean;
+    // Stop accepting new connections.
     public stopListening :() => Promise<void>;
 
-    public shutdown :() => Promise<void>; // stop listening and then close-all
+    // Calls close on all connections. Doesn't stop listening.
+    public closeAll :() => Promise<void>;
+
+    // Stop listening and then close-all
+    public shutdown :() => Promise<void>;
```
trevj, line 45: Suggestion:

``` js
// Stops accepting new connections and then closes all connections.
```

src/tcp/tcp.ts
```diff
@@ -21,14 +20,25 @@ module Tcp {
...
   }
 
-  // A limit on the max number of TCP connections before we start rejecting
-  // new ones.
-  var DEFAULT_MAX_CONNECTIONS = 1048576;
+  // A static helper function to close a freedom socket object and then
+  // appropriately deallocate it's interface object.
+  function closeFreedomSocket_(socket:freedom_TcpSocket.Socket,
+      idInfoString?:string) : Promise<void> {
+    return socket.close().then(() => {
```
trevj: I think you need to a failure handler in case `socket.close` rejects.

src/tcp/tcp.ts
```diff
@@ -21,14 +20,25 @@ module Tcp {
...
+      console.log('closing: ' + idInfoString);
```
trevj: I would suggest removing this logging statement from this method and having the caller log success or failure. The benefit of doing so may be more clear once there are catch blocks in here.

src/tcp/tcp.ts
```diff
@@ -21,14 +20,25 @@ module Tcp {
...
+      freedom['core.tcpsocket'].close(socket);
```
trevj: I'm assuming this is a promise in which case you should return this rather than just calling it.

src/tcp/tcp.ts
```diff
@@ -21,14 +20,25 @@ module Tcp {
...
+      : Net.Endpoint {
      return { address: info.peerAddress, port: info.peerPort }
   }
 
-  // A limit on the max number of TCP connections before we start rejecting
-  // new ones.
-  var DEFAULT_MAX_CONNECTIONS = 1048576;
+  // A static helper function to close a freedom socket object and then
+  // appropriately deallocate it's interface object.
+  function closeFreedomSocket_(socket:freedom_TcpSocket.Socket,
```
trevj: Can you think of a more specific term than "close"?

Also, please move the detailed comment starting line 352 to this function.

src/tcp/tcp.ts
iislucas: Yeah, I thought so at first too, but it is actually not a promise. I checked with Will. It's a synchronous op.

trevj:
So this is looking good. I have a bunch of suggestions but I think this is heading the right direction. Thanks for really digging into the Freedom issues here, especially closing!

While we could add a bunch of unit tests, I suspect they'll end up looking very similar to what I've recently added to the SOCKS server -- ultimately, what we really want for this class are integration tests, which we're not quite ready for yet.

src/tcp/tcp.ts
iislucas: oops, that comment (at 352) is actually done already by this pull request. Removed the comment at 352. Maybe destory is better... using that for now.

src/tcp/tcp.ts
iislucas: Removed all console logs; they were accidentally not removed :) 
I think failure/success should handled by promise fulfillment/rejection.

src/tcp/tcp.ts
iislucas: I think we want failure to be propagated to the caller, so not handled here.

src/tcp/tcp.d.ts
iislucas, line 45: done.

src/tcp/tcp.d.ts
iislucas: done.

src/tcp/tcp.d.ts
iislucas, line 30: done.

src/tcp/tcp.d.ts
iislucas: lol. Done. Thanks.

iislucas:
Done all the one's I think we should & commented on one's I think we don't want. 
Thanks for helpful feedback!
PTAL

src/tcp/tcp.d.ts
trevj, line 30: Don't forget to update the `onceListening` comment as I suggested.

src/tcp/tcp.d.ts
trevj: :+1:

src/tcp/tcp.d.ts
trevj: :+1:

src/tcp/tcp.d.ts
trevj, line 45: :+1:

src/tcp/tcp.ts
trevj: As discussed, I didn't mean we should swallow the exception: we should call `freedom['core.tcpsocket'].close(socket);` even if the call to `socket.close()` rejects (and then reject with the original error).

Like a finally block.

src/tcp/tcp.ts
trevj: Cool. There are still a bunch of commented `log.debug()` statements, though. What do you want to do with them?

src/tcp/tcp.ts
trevj: It would be good to add a comment explaining this because this is different from any Freedom calls I'm familiar with :-)

trevj:
Great. Getting very close now.

src/tcp/tcp.ts
```diff
@@ -21,30 +20,59 @@ module Tcp {
...
+  // Helper function.
+  function endpointOfSocketInfo(info:freedom_TcpSocket.SocketInfo)
+      : Net.Endpoint {
+     return { address: info.peerAddress, port: info.peerPort }
+  }
+
+  // A static helper function to close a freedom socket object and then
+  // appropriately deallocate it's interface object.
+  function destroyFreedomSocket_(socket:freedom_TcpSocket.Socket,
+      idInfoString?:string) : Promise<void> {
```
trevj: You don't need `idInfoString` any more.

src/tcp/tcp.ts
```diff
@@ -21,30 +20,59 @@ module Tcp {
...
   // new ones.
   var DEFAULT_MAX_CONNECTIONS = 1048576;
 
-  // TODO: support starting listening again after stopping
-  // TODO: support changing the connection handler.
-  // TODO: For dynamic port allocation, provide a way to get the post that we
-  // end up listening on.
-  // TODO: make endpoint into getter: we don't support changing it by
-  // assignment.
+  // Helper function.
```
trevj: You can probably remove this comment, it's pretty obvious what this function does.

src/tcp/tcp.ts
trevj: :+1:

trevj:
Just to say that this issue also kinda resolves this:
https://github.com/uProxy/uproxy/issues/468

src/tcp/tcp.d.ts
iislucas, line 30: done.

src/tcp/tcp.ts
iislucas: agreed and done.

src/tcp/tcp.ts
iislucas: Leave them for now. they are helpful for debugging.

src/tcp/tcp.ts
iislucas: done.

iislucas:
PTAL

Thanks!

src/tcp/tcp.d.ts
trevj, line 30: :+1:

src/tcp/tcp.ts
trevj: :+1:

trevj:
Just one thing: you don't need the `idInfoString` argument to `destroyFreedomSocket_` anymore (line 35 of tcp.ts).

Then :+1:

src/tcp/tcp.ts
iislucas: done.

PR #141, opened 2014-10-23
implement similar lifecycle for socks sessions as for socks server

trevj:
Similar ideas to:
https://github.com/uProxy/uproxy-networking/pull/137

A SOCKS session lives as long as its TCP connection and datachannel.

### DIFF
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -26,8 +26,14 @@ declare module SocksToRtc {
     public makeTcpToRtcSession :(tcpConnection:Tcp.Connection) => void;
   }
   class Session {
-    constructor(tcpConnection:Tcp.Connection,
-                peerConnection_:freedom_UproxyPeerConnection.Pc);
+    constructor();
+    public start :(
+        channelLabel:string,
+        tcpConnection:Tcp.Connection,
+        peerConnection_:freedom_UproxyPeerConnection.Pc,
+        bytesSentToPeer:Handler.Queue<number,void>)
+        => Promise<Net.Endpoint>;
+    public stop :() => Promise<void>;
     public tcpConnection :Tcp.Connection;
     public onceReady :Promise<Net.Endpoint>;
     public onceClosed :Promise<void>;
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -36,7 +36,7 @@ declare module SocksToRtc {
     public stop :() => Promise<void>;
     public tcpConnection :Tcp.Connection;
     public onceReady :Promise<Net.Endpoint>;
-    public onceClosed :Promise<void>;
+    public onceStopped :Promise<void>;
     public longId :() => string;
     public close :() => Promise<void>;
     public handleDataFromPeer :(data:WebRtc.Data) => void;
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -38,7 +38,6 @@ declare module SocksToRtc {
     public onceReady :Promise<Net.Endpoint>;
     public onceStopped :Promise<void>;
     public longId :() => string;
-    public close :() => Promise<void>;
     public handleDataFromPeer :(data:WebRtc.Data) => void;
     public channelLabel :() => string;
     public toString :() => string;
--- a/src/socks-to-rtc/socks-to-rtc.spec.ts
+++ b/src/socks-to-rtc/socks-to-rtc.spec.ts
@@ -99,9 +99,6 @@ describe("SOCKS session", function() {
   });
 
   it('onceReady fulfills with listening endpoint on successful negotiation', (done) => {
-    mockTcpConnection.onceClosed = new Promise<Tcp.SocketCloseKind>((F, R) => {});
-    (<any>mockPeerconnection.onceDataChannelClosed).and.returnValue(new Promise<void>((F, R) => {}));
-
     spyOn(session, 'doAuthHandshake_').and.returnValue(Promise.resolve());
     spyOn(session, 'doRequestHandshake_').and.returnValue(Promise.resolve(mockEndpoint));
 
--- a/src/socks-to-rtc/socks-to-rtc.spec.ts
+++ b/src/socks-to-rtc/socks-to-rtc.spec.ts
@@ -6,12 +6,12 @@
 /// <reference path='../webrtc/datachannel.d.ts' />
 /// <reference path='../webrtc/peerconnection.d.ts' />
 
-describe("socksToRtc", function() {
-  var mockEndpoint :Net.Endpoint = {
-    address: '127.0.0.1',
-    port: 1234
-  };
+var mockEndpoint :Net.Endpoint = {
+  address: '127.0.0.1',
+  port: 1234
+};
 
+describe('SOCKS server', function() {
   var server :SocksToRtc.SocksToRtc;
 
   var mockTcpServer :Tcp.Server;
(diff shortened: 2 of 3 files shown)


trevj:
@iislucas since you reviewed the previous pull request I guess you're the best person to look at this (sorry).

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -154,14 +155,39 @@ module SocksToRtc {
...
     public makeTcpToRtcSession = (tcpConnection:Tcp.Connection) : void => {
-      var session = new Session(tcpConnection, this.peerConnection_,
-        this.bytesReceivedFromPeer, this.bytesSentToPeer);
-      this.sessions_[session.channelLabel()] = session;
-      session.onceClosed.then(() => {
-        delete this.sessions_[session.channelLabel()];
+      var tag = obtainTag();
+      log.debug('allocated tag ' + tag + ' for new SOCKS session');
+
+      this.peerConnection_.openDataChannel(tag).then(() => {
```
iislucas, line 172: add a TODO: when we move to using the real peer-connection class (instead of uproxypeerconnection freedom interface), then we can pass the data channel to the session.start function and avoid the peer-connection or the tag ever being needed explicitly in the session. That's going to make things a lot better, both for testing and for code readability.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -211,89 +237,92 @@ module SocksToRtc {
...
 
-    // These are used to avoid double-closure of data channels. We don't need
-    // this for tcp connections because that class already holds the open/
-    // closed state.
-    private dataChannelIsClosed_ :boolean;
+    // Call this to initiate shutdown.
+    private initiateShutdown_ :() => void;
+    private onceStopping_ = new Promise((F, R) => {
+      this.initiateShutdown_ = F;
+    });
```
iislucas, line 258: I think we'd be better to stick with the convention of naming the fulfill promise for a promise named `once<PROMISENAME>` as `fulfill<PROMISENAME>`. Otherwise it makes you think something special is happening.  i.e. rename `initiateShutdown_` to `fulfillShutdown`

Also we should rename the `onceClosed` to `onceShutdown` to mirror your renaming of `close` to `shutdown`, and merge `onceStopping_` and `onceClosed` (there's only a reason for a `onceStopping` if that is public. If it's private it plays no useful role that I can see).

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -211,89 +237,92 @@ module SocksToRtc {
...
+
+    // Fulfills once the SOCKS session has terminated and the TCP connection
+    // and datachannel have been shutdown.
+    // This can happen in response to:
+    //  - startup (negotiation) failure
+    //  - TCP connection or datachannel termination
+    //  - manual invocation of stop()
+    // Rejects if there's an error shutting down the TCP connection or
```
iislucas: closing peer-connection and tcp can never fail; so we comment that this onceClosed also always fulfills. Bonus: then we don't have to deal with failure's we don't know what to do with. That makes less error handling code, but is worth a comment. Ideally we'd have smarter typesystem to say that this is a fulfill only promise.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -211,89 +237,92 @@ module SocksToRtc {
...
+        peerConnection:freedom_UproxyPeerConnection.Pc,
+        bytesSentToPeer:Handler.Queue<number,void>)
+        : Promise<Net.Endpoint> => {
+      this.channelLabel_ = channelLabel;
+      this.tcpConnection_ = tcpConnection;
+      this.peerConnection_ = peerConnection;
+      this.bytesSentToPeer_ = bytesSentToPeer;
+
+      // Startup notifications.
+      this.onceReady = this.doAuthHandshake_().then(this.doRequestHandshake_);
```
iislucas, line 293: I think you need to include the peer-connection onceReady promise in between the above like it was before or else there is a race condition that you might start an auth handshake too soon.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -211,89 +237,92 @@ module SocksToRtc {
...
+    // Initiates shutdown of the TCP server and peerconnection.
+    // Returns onceClosed.
+    public stop = () : Promise<void> => {
+      this.initiateShutdown_();
       return this.onceClosed;
     }
 
+    // Shuts down the TCP server and datachannel.
+    private shutdown_ = () : Promise<void> => {
+      return Promise.all<any>([
```
iislucas: I think you should avoid double-closures and use the boolean-guarded checks for closure of the underlying tcp and peer-connections.

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -83,3 +83,84 @@ describe("socksToRtc", function() {
...
+        'onceDataChannelClosed',
+        'closeDataChannel'
+      ]);
+    mockBytesSent = jasmine.createSpyObj('bytes sent handler', [
+        'handle'
+      ]);
+  });
+
+  it('onceReady fulfills with listening endpoint on successful negotiation', (done) => {
+    mockTcpConnection.onceClosed = new Promise<Tcp.SocketCloseKind>((F, R) => {});
```
iislucas: Do you think it would be worthwhile making the mock objects have names? Are they likely to be useful re-usable things? I actually don't understand why onceClosed is set here...? I don't see it used in this test... :/

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -83,3 +83,84 @@ describe("socksToRtc", function() {
...
+    (<any>mockPeerconnection.onceDataChannelClosed).and.returnValue(new Promise<void>((F, R) => {}));
+
+    spyOn(session, 'doAuthHandshake_').and.returnValue(Promise.resolve());
```
iislucas, line 126: Not sure what I think of spying on internal values... I think we'd be better mocking the tcp and peer-connection for some realistic values and then test the specification in terms of our function call-interactions and IO. Unit tests for internal functions are of slightly dubious value in my mind... but I'm a little undecided and it's late... lets chat a bit tomorrow? Would love to know what you think. If something is so complex it needs to be tested independently, I'd be inclined to think about factoring it our into its own thing... thoughts?

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -83,3 +83,84 @@ describe("socksToRtc", function() {
...
+    spyOn(session, 'doRequestHandshake_').and.returnValue(Promise.resolve(mockEndpoint));
+
+    session.start('buzz', mockTcpConnection, mockPeerconnection, mockBytesSent)
+      .then(session.stop).then(() => { return session.onceClosed; }).then(done);
+  });
+
+  it('onceClosed rejects on tcp connection shutdown failure', (done) => {
```
iislucas: I didn't think tcp connection closure can fail, can it?

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -211,89 +237,92 @@ module SocksToRtc {
...
-      // Note: closing the tcp connection should raise an event to close the
-      // data channel. But we can start closing it down now anyway (faster,
-      // more readable code).
-      if(!this.dataChannelIsClosed_) {
-        this.peerConnection_.closeDataChannel(this.channelLabel_);
-        this.dataChannelIsClosed_ = true;
-      }
+    // Initiates shutdown of the TCP server and peerconnection.
+    // Returns onceClosed.
+    public stop = () : Promise<void> => {
```
iislucas, line 314: I think we can merge the stop and shutdown_ and unify names. :)

src/socks-to-rtc/socks-to-rtc.ts
trevj, line 172: Agreed, and I can't wait for it to happen. There's already a TODO to this effect at line 276.

src/socks-to-rtc/socks-to-rtc.ts
trevj, line 258: Agreed. I've renamed `Session.onceClosed` to `Session.onceStopping`, as per the server. Now  server and session both have:

`onceReady` -> `onceStopping_` -> `onceStopped`

I've also renamed `shutdown_` to `stopResources_` which is more specific and descriptive.

I've found `onceStopping_` useful to provide a hook for external classes to terminate the server or session, in addition to the various ways it can terminate abnormally. Can you think of another way to do that?

src/socks-to-rtc/socks-to-rtc.ts
trevj: That's a good point. I agree this simplifies things. It also lets me remove one unit test.

I'm going to keep the catch block in `makeTcpToRtcSession` to log unforeseen shutdown errors but, otherwise, done (that's a "top of the waterfall" kind of place for uncaught session shutdown errors).

Regarding the server, I looked through the code and it seems neither TCP server nor peerconnection (should) ever reject either. This simplifies a few things and allows me remove one unit test.

What do you think?

src/socks-to-rtc/socks-to-rtc.ts
trevj, line 293: Good catch, however since the datachannel should now be established before this method is called (see `makeTcpToRtcSession`), this is okay.

src/socks-to-rtc/socks-to-rtc.spec.ts
trevj: Correct. See my response above.

src/socks-to-rtc/socks-to-rtc.ts
trevj: As you know, I would like to avoid the required book-keeping if at all possible :-)

Remembering that both SOCKS server and SOCKS session will only call `close` once on each object:
- So, peerconnection is nice. `peerconnection.close` is explicitly idempodent (see its comments). Datachannel closure is idempodent even though the interface doesn't explicitly state it.
- `tcp.ts` is less well defined. `Tcp.Server.close` is not idempodent and has no `isClosed`-type function. We may want to fix that. `Tcp.Connection.close` is logically idempodent but prints a warning if you attempt multiple closures. Gah.

I've updated the comments in the `closeResources_` methods and made a yucky fix for TCP connection closure.

Thoughts?

src/socks-to-rtc/socks-to-rtc.spec.ts
trevj: What do you mean by "have names"? (they do have names -- if you're just asking whether they should be re-useable, my thoughts are _not yet_)

Yeah, actually no need to configure the connection and datachannel closure promise behaviour here. Removed.

src/socks-to-rtc/socks-to-rtc.spec.ts
trevj, line 126: As discussed, I've added TODOs to make `doAuthHandshake_` and `receiveEndpointFromPeer_` public and add tests (kinda redundant since all almost untested functions should have unit tests...but it helps prioritise).

src/socks-to-rtc/socks-to-rtc.ts
trevj, line 314: So, I agree that would be nice.

Can you think of a way to do that and still have this function return `this.onceStopped_`?

trevj:
Thanks again for the detailed review. Just a couple of things bouncing back at you for another look.

src/socks-to-rtc/socks-to-rtc.d.ts
```diff
@@ -26,11 +26,17 @@ declare module SocksToRtc {
...
+        channelLabel:string,
+        tcpConnection:Tcp.Connection,
+        peerConnection_:freedom_UproxyPeerConnection.Pc,
+        bytesSentToPeer:Handler.Queue<number,void>)
+        => Promise<Net.Endpoint>;
+    public stop :() => Promise<void>;
     public tcpConnection :Tcp.Connection;
     public onceReady :Promise<Net.Endpoint>;
-    public onceClosed :Promise<void>;
+    public onceStopped :Promise<void>;
```
iislucas, line 39: Is why rename to onceStopped? Isn't this the promise that happens once close is called? Maybe you'd like to rename close below to stop?

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -92,15 +92,74 @@ describe("socksToRtc", function() {
...
+    spyOn(session, 'doRequestHandshake_').and.returnValue(Promise.reject('unknown hostname'));
+    (<any>mockTcpConnection.isClosed).and.returnValue(false);
+
+    session.start('buzz', mockTcpConnection, mockPeerconnection, mockBytesSent)
+      .catch((e:Error) => { return session.onceStopped; }).then(done);
+  });
+
+  it('onceStopped fulfills on TCP connection termination', (done) => {
+    (<any>mockTcpConnection.onceClosed).and.returnValue(Promise.resolve());
+    (<any>mockPeerconnection.onceDataChannelClosed).and.returnValue(new Promise<void>((F, R) => {}));
```
iislucas: Lets make a global named static constructor for a promise that never fulfills. Then this becomes more readable: 
`(<any>mockPeerconnection.onceDataChannelClosed).and.returnValue(createUnfilfilledPromise());`

But also: let's document why we have this unfilfilled promise here... I don't quite see it's role in the test.

iislucas:
Liking it!  Lets add a synchronous `isStopped`, make an issue for the idea of generalizing query-able promises / destruction promises. And then big fat :+1:

src/socks-to-rtc/socks-to-rtc.spec.ts
trevj: Done.

trevj:
So, I've filed this issue:
https://github.com/uProxy/uproxy/issues/519

And I added code to prevent "double closure" of TCP server. As discussed it's probably overkill but we can re-think later, perhaps with insight from discussion of the above issue.

src/socks-to-rtc/socks-to-rtc.d.ts
iislucas, line 39: ping?

trevj, line 39: Ah, good catch. I did rename `stop` to `close`.

I really don't like maintaining these .d.ts files :-/

iislucas, line 39: Have a spec file, do the SpecDecl thing, get the compiler to check it for you. :)

PR #143, opened 2014-10-27
Create Polymer UI for copypaste sample

lucyhe:


dborkan:
I think the third_party/polymer code should be removed from github and this pull request (updating .gitignore), and instead should be installed using bower - similar to uproxy/uproxy

dborkan:
Same for third_party/i18n if that can also be taken from bower

lucyhe:
I couldn't get i18n with bower, but I made the change for Polymer

src/interfaces/ui-polymer.d.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * ui-polymer.d.ts
```
trevj: Are we likely to need any other `*-polymer.d.ts` files? If not, I'd suggest renaming this `polymer.d.ts`.

.bowerrc
```diff
@@ -0,0 +1,3 @@
+{
+  "directory": "third_party"
```
trevj: So, I guess we don't have a consistent story for where third party stuff should go.

So that we don't have to add keeping editing `.gitignore`, how about we make this `third_party/bower`? (I see that the uproxy repo uses `third_party/lib` which is good for isolation but not very descriptive)

third_party/i18n/i18n_template.js
```diff
@@ -96,7 +96,7 @@ var i18nTemplate = (function() {
   for (var key in handlers) {
     attributeNames.push(key);
   }
-  var selector = '[' + attributeNames.join('],[') + ']';
+  var selector = 'html /deep/ [' + attributeNames.join('], html /deep/ [') + ']';
```
trevj, line 99: If this is a change we're making to the upstream file then it would probably be helpful to mention this in the README.

trevj:
Nice! I don't know Polymer so I've mostly examined this from a build system point of view (that said, I patched this in and the app seems to run fine -- very cool to see translation working).

I've a few suggestions there. One other thing is that while I think _somewhere_ under `src/` probably is the right place for our TypeScript defintions, we've used the term`typings` in a couple of other places so how about that? Also, since I suspect we'll want to move these typings (and the bower and third_party stuff) to uproxy-lib soon, how about `src/networking-typings` to help avoid naming collisions? (that this is kind of a silly name for UX typings suggests they really belong in `uproxy-lib` :-))

src/interfaces/ui-polymer.d.ts
lucyhe: done

.bowerrc
lucyhe: done

third_party/i18n/i18n_template.js
lucyhe, line 99: done

trevj:
:+1:

src/samples/copypaste-socks-chromeapp/root.html
```diff
@@ -0,0 +1,42 @@
...
+    </style>
+    
+    <div><p class='centered'>
+      <select id='languageInput' on-change='{{ updateLanguage }}'>
+        <option value='en' i18n-content='english'></option>
+        <option value='it' i18n-content='italian'></option>
+      </select>        
+    </p></div>
+
+    <copypaste-start id='start' giving='{{ isGiving }}' getting='{{ isGetting }}' hidden?='{{ isGiving || isGetting }}'></copypaste-start>
```
dborkan: It looks to me like isGiving and isGetting are undefined here

src/samples/copypaste-socks-chromeapp/root.html
```diff
@@ -0,0 +1,42 @@
...
+
+    <template if='{{ isGiving }}'>
+      <copypaste-giver id='giver' hidden?='{{ !isGiving }}'></copypaste-giver>
```
dborkan: It seems redundant to have both `if='{{ isGiving }}'`` on the template and`hidden?={{ !isGiving }}``` on the copypaste-giver.  Maybe just keep the hidden?= attribute?

src/samples/copypaste-socks-chromeapp/start.html
```diff
@@ -0,0 +1,43 @@
+<link rel="import" href='lib/third_party/bower/polymer/polymer.html'>
+
+<polymer-element name='copypaste-start' attributes='giving, getting'>
```
dborkan: do giving and getting really need to be attributes?  It seems like you could just default these within start.ts's ready method, and not make the parent set these as HTML attributes.

src/samples/copypaste-socks-chromeapp/start.ts
```diff
@@ -0,0 +1,13 @@
+  Polymer({
+    giveMode: function() {
+    	this.giving = true; 
+    },
+    getMode: function() {
+    	this.getting = true;
+    },
+    ready: function() {
+    	this.giving = false;
```
dborkan: What happens if both this.giving and this.getting are true or both false?  It seems like you just want 1 boolean (or 1 enum)

src/samples/copypaste-socks-chromeapp/getter.ts
```diff
@@ -0,0 +1,25 @@
...
+  parseInboundText: function() {
+  	parsedInboundMessages = parseInboundMessages(this.inboundText);
+  },
+	consumeInboundText: function() {
+		consumeInboundMessage();
+		// Disable the "Start Proxying" button after it's clicked.
+		consumeMessageButton.disabled = true;
+	},
+  ready: function() {
+  	changeLanguage(getBrowserLanguage());
```
dborkan: There is a \t here, can you change to 2 spaces?

src/samples/copypaste-socks-chromeapp/getter.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    step2ContainerNode = this.$.step2ContainerNode;
+    outboundMessageNode = this.$.outboundMessageNode;
+    inboundMessageNode = this.$.inboundMessageNode;    
+    consumeMessageButton = this.$.consumeMessageButton;
+
```
dborkan: Assigning HTML elements to global variables seems like a bad idea to me, some reasons: 
- it means that copypaste-getter is not actually a re-useable component (multiple components will overwrite the same globals)
- it makes code outside this component responsible for updating the HTML display, which would be better encapsulated in this file.

I think a better way to do this is to get rid of these global variables, and use bindings to update the properties you need.  For example with the step2ContainerNode element, you change the style.display property to block (it's initialized as hidden) in popup.js.  Instead it would be better to have some model maintained by popup (could be an object called model like in uProxy, or a bunch of globals, or something passed in as an attribute to coyppaste-getter) - you could then bind HTML in this template to that model.  e.g. in popup.js:

```
var model = { readyForStep2: false };
...
freedom.on('signalForPeer', (signal:WebRtc.SignallingMessage) => {
  model.readyForStep2 = true;
...
```

then in getter.html change this div from:
`<div id='step2ContainerNode' style='display: none;'>`
to:
`<div id='step2ContainerNode' style='display: {{ model.readyForStep2 ? "block" : "none" }};'>`

You can do similar binding for display text, button sensitivity, etc.  Also if you need to change a bunch of style properties you can bind from the model to the class attribute, then use CSS selectors to apply a group of styles at once.

src/samples/copypaste-socks-chromeapp/getter.html
```diff
@@ -0,0 +1,47 @@
...
+        <p class='warning' i18n-content='warning'></p><br>
+        <p i18n-content='pasteAndStartProxying'></p>
+        <textarea id='inboundMessageNode' value='{{ inboundText }}' on-keyup='{{ parseInboundText }}'></textarea>
+        <p>
+          <button id='consumeMessageButton' disabled i18n-content='startProxying' on-click='{{ consumeInboundText }}'></button>
+        </p>
+      </div>
+
+      <div>
+        <a id='receivedBytesNode'>0</a><a i18n-content='bytesReceived'></a><br>
```
dborkan: this could be changed to
`{{ totalBytesReceived }} <span i18n-content='bytesReceived'></span><br>`
(i.e. get rid of all references to receivedBytesNode since you can just update the element using Polymer binding)

dborkan: Also using <a> tags seems a little weird when you aren't using a link, typically we use a div or span tag (they are basically the same except div by default puts the element on a newline while span puts it on the same line, although this isn't always the case depending on css)

dborkan:
:+1:

PR #144, opened 2014-10-28
echo server for firefox

trevj:
All our samples should run on Firefox:
https://github.com/uProxy/uproxy/issues/419

Summary:
- create `src/samples/echo-server-firefoxapp`, with `package.json` and `lib/main.js`
- because the "echo server Freedom module" is common to both Chrome and Firefox, factor it out into its own build component, `echo`

As for all other components, `echo` gets copied into the Chrome and Firefox apps' `lib/` folder from where it can be bootstrapped for each environment.

Tested in Chrome and Firefox.

Thoughts? I don't necessarily love the idea of having the echo server module exist outside of the samples folder but it _is_ common code.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -102,9 +102,13 @@ module.exports = (grunt) ->
       ipaddrjs: Rule.copyModule 'ipaddrjs'
       rtcToNet: Rule.copyModule 'rtc-to-net'
       benchmark: Rule.copyModule 'benchmark'
+      echo: Rule.copyModule 'echo'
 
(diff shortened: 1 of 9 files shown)


salomegeo:
Echo folder should be nested under samples/.

trevj:
Can you think how that would work without making some big changes to our build rules? (I couldn't.)

iislucas:
“Ring the bells that still can ring 
Forget your perfect offering 
There is a crack in everything 
That's how the light gets in.” 
― Leonard Cohen

I think echo in the src dir is ok.

iislucas:
:+1:

PR #146, opened 2014-10-29
SocksToRtc-style lifecycle and unit tests for RtcToNet

trevj:
This pull request is to RtcToNet as the following pull request was to:
https://github.com/uProxy/uproxy-networking/pull/137

Of course, the RtcToNet "server" has only one dependency: the peerconnection. Also remove code related to control channel which is now handled by peerconnection.

Tested with `curl` and a quick browsing session on Firefox.

I will tackle RtcToNet sessions in a followup.


src/rtc-to-net/rtc-to-net.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+// Neither fulfills nor rejects.
+// Useful in a bunch of tests where a promise must be returned
+// for chaining purposes.
+var noopPromise = new Promise<void>((F, R) => {});
+
+describe('module', function() {
+  var server :RtcToNet.RtcToNet;
+
+  // var mockTcpServer :Tcp.Server;
```
iislucas: remove the mock tcp server comment?

src/rtc-to-net/rtc-to-net.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    server.start(mockProxyConfig, mockPeerconnection)
+      .then(() => { return server.onceClosed; })
+      .then(done);
+  });
+
+  it('onceClosed fulfills on call to stop', (done) => {
+    (<any>mockPeerconnection.onceConnected).and.returnValue(Promise.resolve());
+    // peerconnection doesn't close "naturally".
+    (<any>mockPeerconnection.onceDisconnected).and.returnValue(noopPromise);
```
iislucas, line 61: This worries me a little; I would expect onceClosed to complete only after the peer connection is closed, but the mock says peerConnection never closes...

src/rtc-to-net/rtc-to-net.spec.ts
trevj: Oops, done.

src/rtc-to-net/rtc-to-net.spec.ts
trevj, line 61: I want to test that calling `stop()` (and calling `stop()` alone) initiates shutdown and causes `onceClosed` to fulfill.

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -48,20 +46,35 @@ module RtcToNet {
...
+
+    // Fulfills once the module has terminated and the peerconnection has
+    // been shutdown.
+    // This can happen in response to:
+    //  - startup failure
+    //  - peerconnection termination
+    //  - manual invocation of close()
+    // Should never reject.
+    // TODO: rename onceStopped, ala SocksToRtc (API breakage).
     public onceClosed :Promise<void>;
```
iislucas, line 73: Let's make onceClosed and onceReady into function that return the internal value. We don't really want direct external access to these values I think. That will also improve the test-readability: you can write `...then(server.onceClosed)...` ?

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -70,63 +83,83 @@ module RtcToNet {
...
+    // Returns onceClosed.
+    // TODO: rename stop, ala SocksToRtc (API breakage).
+    public close = () : Promise<void> => {
+      this.fulfillStopping_();
+      return this.onceClosed;
+    }
+
+    // Shuts down the peerconnection, fulfilling it has terminated.
+    // Since its close() method should ever reject, this should never reject.
+    public stopResources_ = () => {
```
iislucas: Add the return value's type explicitly to the function type.

src/rtc-to-net/rtc-to-net.ts
trevj, line 73: I completely agree but following up from our conversation last week, how about a followup pull request which makes some of these highly desirable changing API breaks? (it could cover both RtcToNet and SocksToRtc)

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -70,63 +83,83 @@ module RtcToNet {
...
+      // uproxypeerconnection doesn't allow us query whether the
+      // peerconnection has shut down but the call is explicitly idempodent.
+      return this.peerConnection_.close();
```
iislucas, line 163: I'd expect a session close here - why peer connection close?

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -70,63 +83,83 @@ module RtcToNet {
...
       this.peerConnection_.on('signalForPeer', this.signalsForPeer.handle);
+      this.peerConnection_.on('peerOpenedChannel', this.onPeerOpenedChannel_);
+
       this.onceReady = this.peerConnection_.onceConnected().then(() => {});
-      this.onceClosed = this.peerConnection_.onceDisconnected();
-      // TODO: add checking that the peer's fingerprint matches the provided
-      // fingerprint.
+      this.onceReady.catch(this.fulfillStopping_);
+      this.peerConnection_.onceDisconnected()
+          .then(this.fulfillStopping_, this.fulfillStopping_);
```
iislucas, line 121: Shouldn't the fulfill stoppping also close all sessions (else you get left with dangling tcp connections? )

src/rtc-to-net/rtc-to-net.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    server.start(mockProxyConfig, mockPeerconnection)
+      .then(server.close)
+      .then(() => { return server.onceClosed; })
+      .then(done);
+  });
```
iislucas, line 67: Seems like we're missing tests for sessions; but I guess that is for a later pull request (add a TODO?)

iislucas:
Great to see this happening! A few clarifications to make, but pretty close I think.

src/rtc-to-net/rtc-to-net.spec.ts
iislucas, line 61: cool, then let's add a comment so it doesn't look like it might be an oversight, but is intentional.

src/rtc-to-net/rtc-to-net.ts
iislucas, line 73: sounds good.

src/rtc-to-net/rtc-to-net.spec.ts
trevj, line 61: Done.

src/rtc-to-net/rtc-to-net.ts
trevj: Done.

src/rtc-to-net/rtc-to-net.ts
trevj, line 163: RtcToNet is alive so long as the peerconnection is alive. What session do you mean?

src/rtc-to-net/rtc-to-net.ts
trevj, line 121: So, FWIW, SocksToRtc doesn't currently explicitly close its sessions on termination.

For sure, peerconnection should report all its datachannels as closed once the peerconnection closes.

I'm in two minds regarding "second guessing" the peerconnection's behaviour here, and potentially everywhere we use it. That said, perhaps a destructor should code defensively.

Perhaps we should chat about this in person?

src/rtc-to-net/rtc-to-net.spec.ts
trevj, line 67: Yep, for later pull request. I'm already working on it.

src/rtc-to-net/rtc-to-net.ts
iislucas, line 163: sorry, I'm being confusing, I mean I'd expect to see closure of all sessions. TCP resources need cleaning up too.

src/rtc-to-net/rtc-to-net.ts
iislucas, line 121: Yes, let's chat tomorrow.

src/rtc-to-net/rtc-to-net.ts
trevj, line 163: As discussed in person, let's do this here and in SocksToRtc in a followup pull request. I added a TODO and filed an issue:
https://github.com/uProxy/uproxy/issues/561

src/rtc-to-net/rtc-to-net.ts
trevj, line 121: As discussed in person, let's do this here and in SocksToRtc in a followup pull request. I added a TODO and filed an issue:
https://github.com/uProxy/uproxy/issues/561

PR #147, opened 2014-10-30
Support bidirectional text, add Arabic & Farsi translations

lucyhe:
Ran grunt test.

This only adds support for bidirectional text but not layouts. The CSS still needs to be mirrored.


lucyhe:
PTAL
added "addTranslatedStrings" function which only translates the new elements

dborkan:
:+1:

PR #149, opened 2014-11-11
simple socks for firefox

trevj:
So, just like:
https://github.com/uProxy/uproxy-networking/pull/144

This extracts the freedom.js module into a new component, which can be used by both Firefox and Chrome samples.

I'm afraid this doesn't actually make Simple SOCKS work on Firefox because uproxy-core-env doesn't work on Firefox, and that's needed (somehow echo sample works on Firefox...but not this).

However, this pull request should help us with the move to freedom.js v0.6!


trevj:
@bemasc any chance you could take a look at this?

bemasc:
:+1:

PR #150, opened 2014-11-11
Copy/paste sample: Base64-encode signalling messages.

matthild:
This change resolves https://github.com/uProxy/uproxy/issues/562.

### DIFF
--- a/src/samples/copypaste-socks-chromeapp/popup.ts
+++ b/src/samples/copypaste-socks-chromeapp/popup.ts
@@ -14,6 +14,16 @@ var model = { givingOrGetting : <string>null,
               totalBytesSent : 0,
             };
 
+// Define basee64 helper functions that are type-annotated and meaningfully
+// named.
+function base64Encode(unencoded:string): string {
+  return window.btoa(unencoded);
+}
+// Throws an exception if the input is malformed.
+function base64Decode(encoded:string): string {
+  return window.atob(encoded);
+}
+
 // Stores the parsed messages for use later, if & when the user clicks the
 // button for consuming the messages.
 var parsedInboundMessages :WebRtc.SignallingMessage[];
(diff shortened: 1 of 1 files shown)


src/samples/copypaste-socks-chromeapp/popup.ts
```diff
@@ -68,8 +87,14 @@ function consumeInboundMessage() : void {
...
 
-  model.outboundMessageValue =
-      model.outboundMessageValue.trim() + '\n' + JSON.stringify(signal);
+  // Append the new signalling message to the previous message(s), if any.
+  // Base64-encode the concatenated messages because some communication
+  // channels are likely to transform portions of the raw concatenated JSON
+  // into emoticons, whereas the base64 alphabet is much less prone to such
+  // unintended transormation.
+  var oldJson = base64Decode(model.outboundMessageValue.trim());
+  var newJson = oldJson + '\n' + JSON.stringify(signal);
```
salomegeo: is it safe to use '\n' as a splitter? Are we confident that message would never have otherwise?

trevj: Yes, RFC 4566 precludes the use of newlines in SDP messages.

trevj:
Nice! We don't have a good story, longer-term, for how messages should be encoded (for instance, do any social networks do the smiley face conversion? maybe this encoding should live at that layer) but I think is good for now.

The only thing I would suggest is that, if this isn't already the case, the unencoded messages appear in the JavaScript console. They're pretty essential for debugging.

:+1:

matthild:
Thanks for the review.

Yes, the unencoded signalling messages do appear in the Javascript console.  Example log entry:
[PeerConnection](10/11 15:5:4.546) D: rtcNet: handleSignalMessage: 
{"type":2,"candidate":{"candidate":"candidate:895694325 1 udp 2122194687 172.26.93.76 51544 typ host generation 0","sdpMid":"data","sdpMLineIndex":0}}

I've just pushed one additional commit that:
- Fixes a typo in a comment.
- Renames 'oldJson' & 'newJson' to 'oldConcatenatedJson' and 'newConcatenatedJson'.  (These variables' values are not actually JSON, but rather some number of newline-delimited JSON strings concatenated together.)
  Will merge shortly if no objection.

trevj:
Great. Still :+1: !

PR #151, opened 2014-11-11
Copy/paste sample cleanup.

matthild:
- Add missing return value to non-void function.
- Use C++-style comments, not JSDoc-style comments, per uProxy style guide.
- Fix whitespace.


salomegeo:
:+1:

PR #157, opened 2014-11-18
make churn and its samples work with freedom.js v0.6

trevj:
Pretty much the same as the other sample apps.

Depends on:
https://github.com/uProxy/uproxy-lib/pull/90

One thing I noticed about churn is that it didn't like RFC 6544-style "TCP candidates". So I've made `extractEndpointFromCandidateLine` a little more tolerant.


src/churn/churn.ts
```diff
@@ -144,6 +145,11 @@ module Churn {
       this.signalForPeerQueue = new Handler.Queue<Churn.ChurnSignallingMessage,void>();
       this.peerName = config.peerName ||
           'churn-connection-' + crypto.randomUint32();
+
+      // TODO: Remove this public field.
```
bemasc, line 149: Do you feel that way about all the Promises on WebRtc.PeerConnection and WebRtc.DataChannel?  I don't see what's so bad about them.

src/samples/simple-churn-chat-chromeapp/freedom-module.ts
```diff
@@ -72,8 +70,8 @@ b.onceConnected.then(logEndpoints.bind(null, 'b'));
 a.negotiateConnection().then((endpoints:WebRtc.ConnectionAddresses) => {
   a.openDataChannel('text').then((channel:WebRtc.DataChannel) => {
     log.info('datachannel open!');
-	freedom().on('sendA', sendMessage.bind(null, channel));
-	channel.dataFromPeerQueue.setHandler(receiveMessage.bind(null, 'A'));
+  	freedom().on('sendA', sendMessage.bind(null, channel));
```
bemasc: I guess you were fixing indentation ... but now this is 3-space indented?

src/churn/churn.ts
trevj, line 149: Pretty much! I've rarely encountered a good reason not to have a getter (in fact, not sure I've ever seen one).

For example, we have this messiness in some of our tests (a getter function could be supplied to the `then` function directly):
https://github.com/uProxy/uproxy-networking/blob/dev/src/socks-to-rtc/socks-to-rtc.spec.ts#L153

That's separate to the usual encapsulation/immutability arguments.

src/samples/simple-churn-chat-chromeapp/freedom-module.ts
trevj: Oops! A tab crept in. Fixed. Thanks.

src/churn/churn.ts
bemasc, line 149: OK, then I guess you'll want to add an issue like https://github.com/uProxy/uproxy/issues/583.

bemasc:
:+1:

trevj:
Agreed. I made that issue a little more general: just remove public fields.

Thanks for the review!

PR #159, opened 2014-11-18
Unit tests for RtcToNet.Session, with unit tests

trevj:
So, we have a bunch of unit tests for `SocksToRtc.Session` "life-cycle" events, i.e. `onceReady` and `onceStopped`.

This pull request adds a similar set of tests for `RtcToNet.Session`. It also refactors `RtcToNet.Session` to be a little more like `SocksToRtc.Sesson`, chiefly to factor out a separate method, `receiveEndpointFromPeer_`, to deal with SOCKS session handshakes. I think this makes the code much more readable. For example, since there's no longer a single method which receives all data from the peerconnection we can avoid the long and tortuous switch statements exemplified by `handleWebRtcControlMessage_`. It should also be more amenable to testing, e.g. we can add unit tests just for `receiveEndpointFromPeer_`.

`SocksToRtc.Session`'s tests were implemented by this pull request:
https://github.com/uProxy/uproxy-networking/pull/141


bemasc:
The title could use some help.

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -192,161 +192,151 @@ module RtcToNet {
...
-      var request :Socks.Request;
-      try {
-        request = JSON.parse(controlMessage);
-      } catch (e) {
-        log.error(this.longId() + ': Unsupported control message: ' +
-            controlMessage + '; in state: ' +
-            this.toString());
-        return;
+    // Closes the TCP connection and datachannel if they haven't already
+    // closed, fulfilling once both have closed. Since neither objects'
```
bemasc: "object's"

trevj: Oh, I think you're right. Thanks. Fixed here and throughout.

trevj:
As discussed, I re-instated the "bytes received" counter which got left behind in the transition to freedom.js v0.6. I filed an issue to add unit tests for the byte counters:
https://github.com/uProxy/uproxy/issues/605

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -192,161 +188,153 @@ module RtcToNet {
...
+    // Should never reject.
+    private onceStopped_ :Promise<void>;
+    public onceStopped = () : Promise<void> => { return this.onceStopped_; }
 
     // Getters.
     public channelLabel = () : string => { return this.dataChannel_.getLabel(); }
-    public isClosed = () : boolean => { return this.isClosed_; }
 
+    // The supplied datachannel must already be successfully established.
+    // TODO: Rather than passing a reference to the whole peerconnection, we
```
bemasc: This TODO is done...

bemasc:
:+1:

src/rtc-to-net/rtc-to-net.ts
trevj: Removed.

trevj:
Thanks!

PR #160, opened 2014-11-19
Start obfuscated PeerConnection immediately.

bemasc:
This simplifies the code a little.

The simple-chrome-chromeapp sample still works.

### DIFF
--- a/src/churn/churn.ts
+++ b/src/churn/churn.ts
@@ -84,6 +84,8 @@ module Churn {
     public peerName :string;
 
     public onceConnecting :Promise<void>;
+    public onceConnected :Promise<WebRtc.ConnectionAddresses>;
+    public onceDisconnected :Promise<void>;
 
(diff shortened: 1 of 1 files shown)


src/churn/churn.ts
```diff
@@ -239,10 +233,9 @@ module Churn {
       });
     }
 
-    private configureObfuscatedConnection_ = (
-        endpoints:WebRtc.ConnectionAddresses) => {
+    private configureObfuscatedConnection_ = () => {
       log.debug('configuring obfuscated connection...');
```
trevj: Since this is now called from the constructor immediately after configuring the surrogate connection, this is less useful. Perhaps remove this (and the debug line from `configureSurrogateConnection_`)?

trevj:
Nice.

:+1:

src/churn/churn.ts
bemasc: Done.

PR #161, opened 2014-11-19
fixes to allow churn work with the socks samples

trevj:
There were a few bits and pieces missing. This allows churn be used from the samples. Not enabling it by default just yet.


bemasc:
:+1:

PR #162, opened 2014-11-19
sample apps: remove remaining references to DtlsSrtpKeyAgreement and only specify two STUN servers

trevj:
@bemasc I saw you did this for some of the sample apps and both seem like a good idea throughout.


bemasc:
:+1: 

You could even change it to `urls: ['stun:stun.l.google.com:19302', 'stun:stun1.l.google.com:19302']`!

PR #163, opened 2014-11-20
Implement the Thin ICE design in CHURN.

bemasc:


src/churn/churn.ts
```diff
@@ -55,12 +53,77 @@ module Churn {
...
+  }
+
+  export var selectPublicAddress =
+      (candidates:freedom_RTCPeerConnection.RTCIceCandidate[])
+      : NatPair => {
+    var address :string;
+    var port :number;
+    for (var i = 0; i < candidates.length; ++i) {
+      var line = candidates[i].candidate.trim();
+      var tokens = line.split(' ');
```
trevj: It seems like you could use `splitHostCandidateLine_` to split the line and ensure it's a candidate line (and check here that it's a UDP candidate).

src/churn/churn.ts
```diff
@@ -117,25 +195,38 @@ module Churn {
...
+        });
+      });
+
+      // Start the obfuscated connection.
+      this.configureObfuscatedConnection_(config);
+
+      // Once the obfuscated connection's local endpoint is known, the remote
+      // peer has sent us its public endpoint, and probing is complete, we can
+      // configure the obfuscating pipe and allow traffic to flow.
       // the information we need in order to configure the pipes required to
```
trevj: think you can remove this and the next line

trevj:
Very nice. :+1: once you get a chance to address the couple of comments I had.

src/churn/churn.ts
bemasc: Done.

src/churn/churn.ts
bemasc: Done.

PR #165, opened 2014-11-21
remove the TFTP output regex

trevj:
So, this was never meant to be used for very long.

Since we haven't yet decided on what protocol to emulate, I think we would be better off with a cipher.

I believe @kpdyer made some optimisations just before he left which make `.*` equivalent to Rabbit cipher. Let's do that for now.


bemasc:
:+1:

PR #166, opened 2014-11-24
Add source metadata to churn-pipe's Message type.

bemasc:


src/churn-pipe/churn-pipe.ts
```diff
@@ -108,11 +108,23 @@ module Churn {
      * The message is obfuscated before it hits the wire.
      */
     public send = (buffer:ArrayBuffer) => {
+      var remoteEndpoint = {
```
trevj: Suggest you make this a private field which is initialised in the constructor.

src/churn-pipe/churn-pipe.d.ts
```diff
@@ -5,6 +5,7 @@
 declare module freedom_ChurnPipe {
   interface Message {
     data: ArrayBuffer
+    source: Endpoint
```
trevj, line 8: So strictly speaking adding a required field is a breaking API change. Maybe this can be optional, or you can just increase the package number in `package.json`.

trevj:
:+1:

src/churn-pipe/churn-pipe.d.ts
bemasc, line 8: Done.

PR #168, opened 2014-12-12
Make copypaste-churn-chat sample app work.

bemasc:
Note: In my testing, full bidirectional messaging also runs into
https://crbug.com/438699 .

This fixes https://github.com/uProxy/uproxy/issues/652


trevj:
:+1:

PR #170, opened 2014-12-19
Change SocksToRtc to have a freedom-compatible interface.

bemasc:
This change also adds a manifest.json for SocksToRtc, but this
manifest is not yet used anywhere or verified by the build.

### DIFF
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -49,29 +49,28 @@ var rtcNet:RtcToNet.RtcToNet;
 
 freedom().on('start', () => {
   var localhostEndpoint:Net.Endpoint = { address: '127.0.0.1', port: 9999 };
-  socksRtc = new SocksToRtc.SocksToRtc(
-      localhostEndpoint,
-      socksRtcPcConfig,
-      false); // obfuscate
-  log.info('created socks-to-rtc');
+  socksRtc = new SocksToRtc.SocksToRtc();
 
   // Forward signalling channel messages to the UI.
-  socksRtc.signalsForPeer.setSyncHandler((signal:WebRtc.SignallingMessage) => {
+  socksRtc.on('signalForPeer', (signal:any) => {
       freedom().emit('signalForPeer', signal);
   });
 
   // SocksToRtc adds the number of bytes it sends/receives to its respective
   // queue as it proxies. When new numbers (of bytes) are added to these queues,
   // emit the number to the UI (look for corresponding freedom.on in main.html).
-  socksRtc.bytesReceivedFromPeer.setSyncHandler((numBytes:number) => {
+  socksRtc.on('bytesReceivedFromPeer', (numBytes:number) => {
       freedom().emit('bytesReceived', numBytes);
   });
 
-  socksRtc.bytesSentToPeer.setSyncHandler((numBytes:number) => {
+  socksRtc.on('bytesSentToPeer', (numBytes:number) => {
       freedom().emit('bytesSent', numBytes);
   });
 
-  socksRtc.onceReady
+  socksRtc.start(
+      localhostEndpoint,
+      socksRtcPcConfig,
+      false) // obfuscate
     .then((endpoint:Net.Endpoint) => {
       log.info('socksRtc ready. listening to SOCKS5 on: ' + JSON.stringify(endpoint));
       log.info('` curl -x socks5h://localhost:9999 www.google.com `')
@@ -80,6 +79,7 @@ freedom().on('start', () => {
     .catch((e) => {
       console.error('socksRtc Error: ' + e + '; ' + this.socksRtc.toString());
     });
+  log.info('created socks-to-rtc');
 
   socksRtc.onceStopped().then(() => {
     freedom().emit('proxyingStopped');
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -67,6 +67,10 @@ freedom().on('start', () => {
       freedom().emit('bytesSent', numBytes);
   });
 
+  socksRtc.on('stopped', () => {
+    freedom().emit('proxyingStopped');
+  });
+
   socksRtc.start(
       localhostEndpoint,
       socksRtcPcConfig,
@@ -80,10 +84,6 @@ freedom().on('start', () => {
       console.error('socksRtc Error: ' + e + '; ' + this.socksRtc.toString());
     });
   log.info('created socks-to-rtc');
-
-  socksRtc.onceStopped().then(() => {
-    freedom().emit('proxyingStopped');
-  });
 });
 
 // Receive signalling channel messages from the UI.
--- a/src/socks-to-rtc/freedom.json
+++ b/src/socks-to-rtc/freedom.json
@@ -1,0 +1,77 @@
+{
+  "name": "Socks-to-RTC",
+  "description": "SOCKS server that forwards requests to RtcToNet",
+  "app": {
+    "script": [
+      "../socks-common/socks-headers.js",
+      "../logging/logging.js",
+      "../churn/churn.js",
+      "../webrtc/datachannel.js",
+      "../webrtc/peerconnection.js",
+      "../tcp/tcp.js",
+    ]
+  },
+  "provides": [
+    "SocksToRtc"
+  ],
+  "api": {
+    "SocksToRtc": {
+      "start": {
+        "value": [{
+            "address": "string",
+            "port": "number"
+          },
+          {
+            "webrtcPcConfig": {
+              "iceServers": ["array", {
+                "urls": ["array", "string"],
+                "username": "string",
+                "credential": "string"
+              }],
+              "iceTransports": "string",
+              "peerIdentity": "string"
+            },
+            "peerName": "string",
+            "initiateConnection": "boolean"
+          },
+          "boolean"
+        ],
+        "ret": {
+          "address": "string",
+          "port": "number"
+        }
+      },
+      "stop": {
+        "type": "method",
+        "value": []
+      },
+      "onceStopped": {
+        "type": "method",
+        "value": []
+      },
+      "signalForPeer": {
+        "type": "event",
+        "value": "object"
+      },
+      "bytesReceivedFromPeer": {
+        "type": "event",
+        "value": "number"
+      },
+      "bytesSentToPeer": {
+        "type": "event",
+        "value": "number"
+      },
+      "handleSignalFromPeer": {
+        "type": "method",
+        "value": ["object"]
+      }
+    }
+  },
+  "permissions": [
+    "churnPipe",
+    "core.console",
+    "core.rtcdatachannel",
+    "core.rtcpeerconnection",
+    "core.tcpsocket"
+  ]
+}
--- a/src/socks-to-rtc/freedom.json
+++ b/src/socks-to-rtc/freedom.json
@@ -45,8 +45,8 @@
         "type": "method",
         "value": []
       },
-      "onceStopped": {
-        "type": "method",
+      "stopped": {
+        "type": "event",
         "value": []
       },
       "signalForPeer": {
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -6,24 +6,32 @@
 /// <reference path="../third_party/typings/es6-promise/es6-promise.d.ts" />
 declare module SocksToRtc {
   class SocksToRtc {
-    constructor(endpoint?:Net.Endpoint,
-                pcConfig?:WebRtc.PeerConnectionConfig,
-                obfuscate?:boolean);
+    // TODO: make |dispatchEvent| non-optional once modularization is complete.
+    constructor(dispatchEvent?:(t:string, f:Function) => any);
+    // NOTE: The caller MUST set the on-event handlers before calling start().
+    public start :(endpoint:Net.Endpoint,
+                   pcConfig:WebRtc.PeerConnectionConfig,
+                   obfuscate?:boolean) => Promise<Net.Endpoint>;
     public stop :() => Promise<void>;
-    public onceReady :Promise<Net.Endpoint>;
-    public isStopped :() => boolean;
     public onceStopped :() => Promise<void>;
-    public signalsForPeer :Handler.Queue<WebRtc.SignallingMessage, void>;
-    public bytesReceivedFromPeer :Handler.Queue<number, void>;
-    public bytesSentToPeer :Handler.Queue<number, void>;
-    public handleSignalFromPeer :(signal: WebRtc.SignallingMessage) => void;
+
+    on(t:string, f:Function) : void;
+    on(t:'signalForPeer', f:(message:Object) => void) : void;
+    on(t:'bytesReceivedFromPeer', f:(bytes:number) => void) : void;
+    on(t:'bytesSentToPeer', f:(bytes:number) => void) : void;
+    public handleSignalFromPeer :(signal: WebRtc.SignallingMessage) => Promise<void>;
+
+    // These methods are exposed here only for testing purposes, and are not
+    // actually available through the freedom interface.
     public toString :() => string;
-    public start :(
+    public startInternal :(
         tcpServer:Tcp.Server,
         peerconnection:WebRtc.PeerConnection)
         => Promise<Net.Endpoint>;
-    public makeTcpToRtcSession :(tcpConnection:Tcp.Connection) => void;
   }
+
+  // This class is exposed here only for testing purposes, and is not
+  // actually available through the freedom interface.
   class Session {
     constructor();
     public start :(
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -13,9 +13,9 @@ declare module SocksToRtc {
                    pcConfig:WebRtc.PeerConnectionConfig,
                    obfuscate?:boolean) => Promise<Net.Endpoint>;
     public stop :() => Promise<void>;
-    public onceStopped :() => Promise<void>;
 
     on(t:string, f:Function) : void;
+    on(t:'stopped', f:() => void) : void;
     on(t:'signalForPeer', f:(message:Object) => void) : void;
     on(t:'bytesReceivedFromPeer', f:(bytes:number) => void) : void;
     on(t:'bytesSentToPeer', f:(bytes:number) => void) : void;
--- a/src/socks-to-rtc/socks-to-rtc.spec.ts
+++ b/src/socks-to-rtc/socks-to-rtc.spec.ts
@@ -45,6 +45,7 @@ describe('SOCKS server', function() {
 
     mockPeerConnection = <any>{
       dataChannels: {},
+      signalForPeerQueue: new Handler.Queue<WebRtc.SignallingMessage, void>(),
       negotiateConnection: jasmine.createSpy('negotiateConnection'),
       onceConnecting: noopPromise,
       onceConnected: noopPromise,
@@ -59,7 +60,7 @@ describe('SOCKS server', function() {
     // We're not testing termination.
     (<any>mockTcpServer.onceShutdown).and.returnValue(noopPromise);
 
-    server.start(mockTcpServer, mockPeerConnection)
+    server.startInternal(mockTcpServer, mockPeerConnection)
       .then((result:Net.Endpoint) => {
         expect(result.address).toEqual(mockEndpoint.address);
         expect(result.port).toEqual(mockEndpoint.port);
@@ -72,7 +73,7 @@ describe('SOCKS server', function() {
         .and.returnValue(Promise.reject(new Error('could not allocate port')));
     (<any>mockTcpServer.onceShutdown).and.returnValue(Promise.resolve());
 
-    server.start(mockTcpServer, mockPeerConnection).catch(server.onceStopped).then(done);
+    server.startInternal(mockTcpServer, mockPeerConnection).catch(server.onceStopped).then(done);
   });
 
   it('onceStopped fulfills on peerconnection termination', (done) => {
@@ -81,7 +82,7 @@ describe('SOCKS server', function() {
     mockPeerConnection.onceConnected = voidPromise;
     mockPeerConnection.onceDisconnected = <any>(Promise.resolve());
 
-    server.start(mockTcpServer, mockPeerConnection).then(server.onceStopped).then(done);
+    server.startInternal(mockTcpServer, mockPeerConnection).then(server.onceStopped).then(done);
   });
 
   it('onceStopped fulfills on call to stop', (done) => {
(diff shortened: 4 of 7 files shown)


bemasc:
@willscott @trevj @iislucas, PTAL.

My roadmap is:
1. (this change) Change SocksToRtc to have a freedom-compatible interface.
1.1. npm publish.
2. Change uProxy to use SocksToRtc with the new interface.
3. Change uProxy to use SocksToRtc in a freedom module
- May require changes to the uproxy-networking gruntfile.

4-6. Repeat above for RtcToNet.

src/samples/copypaste-socks-chromeapp/freedom-module.ts
```diff
@@ -49,29 +49,28 @@ var rtcNet:RtcToNet.RtcToNet;
...
-  socksRtc = new SocksToRtc.SocksToRtc(
-      localhostEndpoint,
-      socksRtcPcConfig,
-      false); // obfuscate
-  log.info('created socks-to-rtc');
+  socksRtc = new SocksToRtc.SocksToRtc();
 
   // Forward signalling channel messages to the UI.
-  socksRtc.signalsForPeer.setSyncHandler((signal:WebRtc.SignallingMessage) => {
+  socksRtc.on('signalForPeer', (signal:any) => {
```
iislucas, line 55: why is the signal now `any` when it was `WebRtc.SignallingMessage` before?

src/socks-to-rtc/socks-to-rtc.d.ts
```diff
@@ -6,24 +6,32 @@
...
-    public onceReady :Promise<Net.Endpoint>;
-    public isStopped :() => boolean;
     public onceStopped :() => Promise<void>;
-    public signalsForPeer :Handler.Queue<WebRtc.SignallingMessage, void>;
-    public bytesReceivedFromPeer :Handler.Queue<number, void>;
-    public bytesSentToPeer :Handler.Queue<number, void>;
-    public handleSignalFromPeer :(signal: WebRtc.SignallingMessage) => void;
+
+    on(t:string, f:Function) : void;
+    on(t:'signalForPeer', f:(message:Object) => void) : void;
```
iislucas, line 19: Why are the signals for the peer not typed as `WebRtc.SignallingMessage` ?

src/socks-to-rtc/freedom.json
```diff
@@ -0,0 +1,77 @@
...
+          "address": "string",
+          "port": "number"
+        }
+      },
+      "stop": {
+        "type": "method",
+        "value": []
+      },
+      "onceStopped": {
+        "type": "method",
```
willscott: could this be an event?
`socksRTC.once('stopped', onStopHandler);`

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -152,6 +170,21 @@ module SocksToRtc {
       return this.onceStopped_;
     }
 
+    // An implementation of dispatchEvent to use if none has been provided
+    // (i.e. when this class is not being used as a freedom    // module).
+    // For simplicity, only one listener per message type is supported.
+    private fallbackDispatchEvent_ = (t:string, msg:any) : void => {
```
willscott, line 176: I get that this shim is temporary, but can can it live as a separate utility mixin so that it can be applied to other dual-purpose classes in the future?

willscott:
Looks good. Main design question is whether `onceStopped` would work better as a `stopped` event, otherwise very happy with everything here.

src/samples/copypaste-socks-chromeapp/freedom-module.ts
bemasc, line 55: In fact, the previous annotation was incorrect.  When CHURN is enabled, this message is not a SignallingMessage.  However, typescript's typechecking is not strict enough to detect this typing error.

I have not cleaned up every instance of confused typing, but I have tried to remove them from lines that I was already editing.

src/socks-to-rtc/socks-to-rtc.d.ts
bemasc, line 19: In obfuscated mode, the signals have a different JSON structure.

src/socks-to-rtc/freedom.json
bemasc: Done.

willscott:
:+1: from me

src/socks-to-rtc/socks-to-rtc.ts
bemasc, line 176: We spent a bunch of time today talking about this.  I think it's a good idea, and we should try something along those lines, but it looks sufficiently far from trivial that it's better off in a separate pull request.

src/samples/copypaste-socks-chromeapp/freedom-module.ts
iislucas, line 55: This doesn't sound right to me. Can you explain? It's really easy to get these types wrong and the old type did previously catch a bunch of errors. I don't see what the problem with giving this the right type is... ?

iislucas:
If something can only happen once (like stopping) then it's best to use the promise-style. It provides better way's to chain actions after the singleton-event has happened.

iislucas:
I think this was merged prematurely; @trevj and I hadn't finished reviewing. You may need to do a little followup.

src/socks-to-rtc/socks-to-rtc.d.ts
iislucas, line 19: You should make it have a common structure; else you can't tell how you have to treat it on the consumer side.

src/socks-to-rtc/freedom.json
iislucas: Can it really happen more than once?

iislucas:
Please add to your description how this change was tested. (our style is to say how pull requests were tested so that people don't forget, and so that others later on building on earlier work can get hints on how to test their new change.)

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -162,7 +169,7 @@ describe("SOCKS session", function() {
 
   it('onceStopped fulfills on call to stop', (done) => {
     // Neither TCP connection nor datachannel close "naturally".
-    mockTcpConnection.onceClosed = noopPromise;
+    mockTcpConnection.onceClosed = new Promise<Tcp.SocketCloseKind>((F, R) => {});
```
iislucas, line 172: Consider giving noopPromise a type-parameter?

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -71,6 +63,15 @@ module SocksToRtc {
...
     private peerConnection_  :WebRtc.PeerConnection;
 
+    // Event listener registration function.  When running in freedom, this is
+    // not defined, and the corresponding functionality is inserted by freedom
+    // on the consumer side.
+    public on : (t:string, f:(m:any) => void) => void;
+    // Database of event listeners for fallback implementation of |on|.
+    private listeners_ : { [s:string]: (m:any) => void };
+    // TODO: Remove |on| and |listeners_| once all users of this class do so
+    // via freedom.
```
iislucas, line 73: what does the `do so` refer to ? Elaborate comment a bit more.

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -81,23 +82,33 @@ module SocksToRtc {
...
-    // As configure() but handles creation of a TCP server and peerconnection.
-    constructor(
-        endpoint?:Net.Endpoint,
-        pcConfig?:WebRtc.PeerConnectionConfig,
-        obfuscate?:boolean) {
-      if (endpoint) {
-        this.start(
-            new Tcp.Server(endpoint),
-            obfuscate ?
+    constructor(private dispatchEvent_?:(t:string, m:any) => any) {
```
iislucas, line 85: DispatchEvent returns void. You should not let editors in the future use the return value of dispatch event. :)

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -81,23 +82,33 @@ module SocksToRtc {
...
+      if (!this.dispatchEvent_) {
+        // TODO: Remove this code once all users of this class move to freedom.
```
iislucas, line 87: Unclear to me this would be better... unit tests are easier if you don't have to load as much stuff. So it may be that we'd always want this. Propose: change TODO to CONSIDER.

iislucas:
good stuff; only real thing is that I'm not sure about the stop event vs onceStopped promise. I think it's much better to have a onceStopped promise if it only happens once (which was the original design, and still seems like the current design).

src/samples/copypaste-socks-chromeapp/freedom-module.ts
bemasc, line 55: Technically, the correct type here is something like "WebRtc.SignallingMessage OR Churn.ChurnSignallingMessage", because either kind may be passed, depending on the value of the parameter `obfuscated` that is passed to `start()`.  Previously, ChurnSignallingMessage was a subtype of SignallingMessage, but as of https://github.com/uProxy/uproxy-networking/commit/069602034212daff969f5028544876831ee44df0 that is no longer true, and the two interfaces have no non-trivial shared supertype.

The good news is that these sorts of [union types supposedly will be supported](http://blogs.msdn.com/b/typescript/archive/2014/11/18/what-s-new-in-the-typescript-type-system.aspx) in Typescript 1.4.

bemasc:
@willscott and @iislucas, please reach consensus on onceStopped() vs. on('stopped', ...)

ryscheng:
We are already emitting a bunch of events. I think on('stopped', ...) makes
more sense here

src/socks-to-rtc/socks-to-rtc.spec.ts
bemasc, line 172: noopPromise is just a variable, so I'm pretty sure it can't have a type-parameter.  Are you suggesting a factory pattern, something like

``` typescript
function noopPromise<T>() {
  return new Promise<T>((F, R) => {});
}
```

and then replacing every instance of `noopPromise` with `noopPromise<void>()` or `noopPromise<Tcp.SocketCloseKind>()`?

src/socks-to-rtc/socks-to-rtc.ts
bemasc, line 73: Done.  This pull request is closed, so to see the change you'll have to go to https://github.com/uProxy/uproxy-networking/pull/171

src/socks-to-rtc/socks-to-rtc.ts
bemasc, line 85: Done.  This pull request is closed, so to see the change you'll have to go to https://github.com/uProxy/uproxy-networking/pull/171

src/socks-to-rtc/socks-to-rtc.ts
bemasc, line 87: Done.  This pull request is closed, so to see the change you'll have to go to https://github.com/uProxy/uproxy-networking/pull/171

willscott:
@iislucas, why is a `onceStopped` method preferable to attaching a handler to listen for one instance of an event like:

``` javascript
socksToRtc.once('stopped', myOnStoppedHandler);
```

The latter sends one message, instead of 2 per handler registered, and doesn't require the producing class to manage promises for tracking each close handler attached to it.

src/samples/copypaste-socks-chromeapp/freedom-module.ts
iislucas, line 55: Tagged unions are what we want (which we mimic with optional types for alternative possible properties of an object, and assume only one of them is present at a time; support for this is also being discussed). Ordinary unions are really for cases where someone has written a badly defined interface and you have to live with it. 

We already do use a tagged union in the `WebRtc.SignallingMessage` type for the different kinds of signaling messages that are possible. This helps make explicit when you access a particular (sub)type of a tagged object. 

As per our style guide, use of `any` is strongly discouraged as it gives you no type-safety and results in lots of debugging. Of course you could write tests for each type-check condition, but instead we use the type system for this kind of basic stuff (there was a time where we had a bunch of type-checking tests and it was really gross), and so we can focus on writing tests for the stuff that is more important and harder to be done automatically. 

Suggested fix here is that you create a tagged union type and make the signal have a well-defined interface - avoiding I hope laborious future debugging (I've seen several people debug type bugs on this object previously, and would love to avoid it again). 

The current use of lenient rules is very bad and we have an issue to fix it (https://github.com/uProxy/uproxy/issues/681), and I will make that happen soon. Else we're shooting outself in the foot again and I don't want to see that happen again! :) 

Make sense? I think it might be reasonable to do that as a separate pull request as it looks like the interface definition has already been broken, so it is indeed an authogonal issue to this pull request. 

Sound good?

iislucas:
The 2 messages is an implementation choice on Freedom's part; first class support for promises/futures in interfaces would make this a single message. But either way, in this case, performance is not relevant, but of course I would love to see improved support in freedom in the future. 

Three things I like about promises: 

 1 \* From a programming perspective, I prefer promises to attaching singleton function event handlers as it directly support several things being done after event happens (in this case, I can add references in several places in the code for stuff to be done after the stopping happens without worrying about clobbering the existing event handler - I've seen this bug happen a few times in uProxy code base and more generally too). 

 2 \* From a code comprehensibility perspective, I prefer promises as they are guaranteed to only happen once. Using events gives the impression that something may happen many times. Socks-rtc object has a simple life-cyle and it's better if that's immediately obvious from the interface. 

 3 \* Finally, when you use promises, you don't have to worry about when you attach the handler which avoids a class of race-conditions that I've seen people write on and off. 

The general downside with using promises is (4) that coders have to understand them at least a bit. But it is pretty clear that anyone working on uProxy has to understand promises already, so I don't see that as a problem in this case.

willscott:
I guess what I'm wondering if you feel like you don't get points 1 and 2 at
least from the `.once` event handler. It also guarantees execution only
once, and as many handlers as desired can be added to the queue without
danger of clobbering each other.

It sounds like you're imagining a promise 'property' on an object - is that
the right mental model for the envisioned design?

iislucas:
I forgot about the special `.once` event handlers... can you point to an example of their use to remind us how they work (I remember them, but I think most people won't have seen it before, and I forget their syntax implications...) ?

Yes, I was thinking of something like a getter for a promise property. What do you think of that replacing the `.once` handler?

Thanks!

willscott:
`.once` is listed as a helper method included in on/emit interfaces in
our documentation:
https://github.com/freedomjs/freedom/wiki/freedom.js-structure:-Consumer-Interface#the-null-interface

The convention for `.once` seems to have some standardization across
frameworks:
http://nodejs.org/api/events.html#events_emitter_once_event_listener
http://backbonejs.org/#Events-once
http://api.jquery.com/one/

we use it regularly through freedom code:
https://github.com/freedomjs/freedom/blob/master/src/link/worker.js#L116
https://github.com/freedomjs/freedom/blob/master/providers/core/core.peerconnection.js#L362

Promise getters seem further from the 'eventHandler' model that on-emit is
modeled on, so I'm not really convinced about wanting to switch freedom to
expose events through a promise - because I feel like that would be an even
more unexpected interface for someone coming in fresh than what we have now.

iislucas:
Looks to me like the notion of `once` is attaching a one-time listener, but that doesn't say anything about the emitter only having a singleton event emitted. A single event being emitted is the semantics we actually want in this case, and which is what the promise gives. So I guess we're really talking about two different things that have some overlap (namely on point 1, it looks like...). 

For point 2, reading around, sounds like `once` can be seen as an abbreviation as follows `x.once(e,f)`:

```
f_once_handler = (y) => { x. removeListener(e,f_once_handler); f(y); };
x.on(e,f_once_handler);
```

In which case it clobbers other handlers. But maybe (from your comment above) it's a special multiple-event thing where each event handler gets called once independently of other on handlers for the event? ... sound to me from Freedom docs like it's the former, in which case it doesn't help with the syntax issue in (2). Might be good to clarify in the docs what the semantics is if you have both once and on, or two calls to once. 

So, if `once` is what I think it is in Freedom, then `Promises` and `once` are useful for largely separate cases.

Sound like the correct analysis to you?

willscott:
I agree with code of what `once` is an abbreviation for.

Handlers are independent.  The `removeListener` only removes this
specific handler, it does not prevent other handlers from running. All
registered `on` or `once` handlers will run for an incoming event.
Is that what you're worried about?

(this is the convention across just about all eventHandler interface.
Anywhere you call `addEventListener` you should expect that behavior.
It's only when you explicitly over-write a singleton handler
`element.onClick = myHandler` that you clobber others.  The clobbering
style has largely been deprecated through JS interfaces since it's more
brittle, the exception being some of the chrome APIs which still use the
legacy format, and WebRTC, for some inexplicable reason.)

iislucas:
Oh, I didn't know freedom 0.6 uses event listeners and has `removeListener` and that the `on` method in fact adds an event listener instead of sets the event listening function... If that's the case, it would be good to add to the documentation in https://github.com/freedomjs/freedom/wiki/freedom.js-structure:-Consumer-Interface  If so, then I agree that is the standard way to deal with (1) for event-handling - I just wasn't aware freedom had that (A search for `removeListener` doesn't show any results... https://github.com/freedomjs/freedom/search?utf8=%E2%9C%93&q=removeListener so maybe it doesn't? If not then I guess we should add an issue for it?

willscott:
Removing a handler is done with `off` (the example just above `once` on that page).

Off is trying to match `on`. on/off seem to be the less-verbose version of `addEventListener`/`removeEventListener`.
http://www.w3schools.com/jquery/event_off.asp
http://backbonejs.org/#Events-off

willscott:
Also, maybe worth mentioning that this is not something that's changed in 0.6.  `on`,`off``,``once`` are how events are managed in freedom 0.5 as well.

iislucas:
nice! 

(and sorry, I had blinkers on looking for `remove*` as that's what I've seen other places; just goes to show much of a role common language idioms play. I would add briefly to the docs that off is like removeListener and on and once are both forms of addListener so that others doing similar searches find what they are looking for. )

Seems like we agree then that only points (2) and (3) are different w.r.t. promises.

willscott:
I don't see there being a huge difference on (2) - the point where I'd like the guarantee that my handler gets called once is on the consumer side, right?  Using `.once` seems pretty clear that that will happen, as does calling a method, which also is understood to have that property.

From the producer side, neither incarnation guarantees a single point of 'stopping' - the stop event can be emitted multiple times, or the promise chain of called methods can get reset.

The point in (3) is that you'd like to be able to check if an object has reached some state post-facto? I'm not sure if the closed case here is applicable, since the general presumption with objects that are 'dead' is that they should be fully de-referenced at that point, and shouldn't have any further methods called on them - since the backing object needs to be able to be garbage collected.

iislucas:
That's not the way I think of (2) and (3). 

For (2): I meant comprehensibility of the interface and code that uses it. The intention is to express lifecycle pattern for the object in question in the interface and in its usage (in this case socks-rtc). If it can be started and stopped again, then it makes sense to have start method and a stop event (and comments to express if two calls to `start` and two calls to stop are errors or idempotent). But if the object is intended to live, stop, and then be discarded (perhaps after being examined), then having a stop event is suggesting that stop event can happen multiple times, and that a consumer should be prepared to handle these events. You can write a comment to explain when stop events should be expected, and to clarify that it only happens once. But if `stop` is something that can happen only once, then a promise makes that clear in the interface, avoids bad consumer usage patterns, and you get the code to explain itself. When you read the consumer code, you don't have to worry about thinking "wait, what if it the stop event happens again later? did the writer of this code miss that?". 

The general rule is to make your interface express as clearly as possible the behaviour, as opposed to having to explain all the semantics in comments. e.g. going in the opposite direction, and making the interface less informative (but more concise and uniform!) you could have every interface as simply be `{ do(arg:any) : Promise<any> }`. But of course, then the interface isn't telling you much about the intended semantics! So, in generally, prefer to be more precise and have your interface give more detail about the behaviour. 

w.r.t. (3) objects being dead and closed are separate things (especially if you think that it can be started and stopped again in the future); closed objects can be examined. Garbage collected objects cannot. More generally, things get surprisingly complicated with module boundaries and associated resources (e.g. sockets) and undefined time gaps between closure, reacting to the closure, and acting on results: you get all sorts of new classes of race conditions. Having a promise for a one-time state transition is a good habit as it lets you not worry about missing the transition. E.g. we saw some hard to figure out 'errors' with TCP sockets being created, but at the callback for the creation, we found that they had in fact already been closed and discarded: module boundaries make life-cycle of remote objects harder to understand, and it makes it harder to track object existence between module boundaries. So, the simpler the model and the more precise you can be about when things happen, the better: fewer cases of worrying about when you have to handle things, and less debugging of race conditions. The move to a `promisified` interface for socks-rtc fixed a lot of bugs, and I'd like to avoid re-introducing them as they were hard to find and fix.  In the specific case of (3) for socks-rtc as a freedom module, as it's a remote module class, you don't get garbage collection done for you: you have to dispose of it manually. So, separating closure helps the consumer decide when that explicit garbage step happens, and if they choose, do some object introspection before closure happens. But mostly, it's just about good style here and not making readers of the code have to ask so many questions when reading the code. At least, that's the way it looks to me.

willscott:
Lets set up time to chat about this - maybe after the holidays :smile:

I think I see where you're coming from on number two - but I don't think we get convention, conciseness, or 1-time guarantees from promises:
1. A singleton event is not unprecedented, and using events for close is pretty standard. They're used by [node streams](http://nodejs.org/api/stream.html#stream_event_end), [backbone's destroy](http://backbonejs.org/#Model-destroy), and HTML window events like `load` and `close`.
2. Events take less code - the producer just `dispatchEvent('stop');`'s, as opposed to having a method and singleton promise that gets returned. On the consumer side, you can replicate the functionality in one line if desired:
   
   ``` javascript
   obj.onceStopped = new Promise(obj.once.bind(obj, 'stop'));
   ```
3. You don't know if the Promise's that you're getting from different calls to the method will all be the same, or if once the provider has closed subsequent calls won't return, so the same concerns about missing the state change / having the state change happen multiple times arise if you don't look at the implementation / comments.

willscott:
@iislucas - I want to make sure you're okay with the the state of the world we have now with `stopped` implemented as an event.  We should set up a meeting next week if you feel strongly that this isn't a good implementation path, and find a compromise we're both happy with.

trevj:
`onceStopped`

trevj:
:+1:, so far as modularising `SocksToRtc` goes.

`onceStopped` worked fine for us and I was happy with how we could use promise chaining in our tests, e.g.:

``` typescript
server.startInternal(mockTcpServer, mockPeerConnection)
    .catch(server.onceStopped).then(done);
```

Which has now become:

``` typescript
var serverStopped = new Promise<void>((F, R) => {
  server.on('stopped', F);
});
onceServerStopped = () => { return serverStopped; };

server.startInternal(mockTcpServer, mockPeerConnection)
    .catch(onceServerStopped).then(done);
```

Not the end of the world but I'm dismayed to see so much time spent discussing the API change. I thought this modularisation was motivated by sandboxing?

willscott:
I guess it's a matter of style, but when you return the onceStopped promise to `catch` you end up with mismatched types, since cught functions are supposed handle an `Error` object, which onceStopped isn't really typed for.  Also, doesn't the example just become:

``` javascript
server.startInternal(mockTcpServer, mockPeerConnection).then(<On Started>)
server.once('stopped', done);
```

It seems like you may well want to have your promise chain let you handle the start call completing, which you don't get if you directly chain the stopped event.

PR #172, opened 2014-12-30
Handle error in tcp socket connection. 

quantsword:
Handle error in tcp socket connection. When it happens, close the tcp connection so that
other code waiting on the pipe can take actions.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/172)

### DIFF
--- a/src/tcp/tcp.ts
+++ b/src/tcp/tcp.ts
@@ -300,7 +300,9 @@ module Tcp {
               log.error('Failed to connect to %1, err=%2',
                         [JSON.stringify(connectionKind.endpoint),
                          JSON.stringify(err)]);
-              this.close();
+              this.dataToSocketQueue.stopHandling();
+              this.state_ = Connection.State.CLOSED;
+              this.fulfillClosed_(SocketCloseKind.NEVER_CONNECTED);
             });
       } else {
         throw(new Error(this.connectionId +
--- a/src/tcp/tcp.ts
+++ b/src/tcp/tcp.ts
@@ -285,7 +285,7 @@ module Tcp {
                 .connect(connectionKind.endpoint.address,
                          connectionKind.endpoint.port)
                 .then(this.connectionSocket_.getInfo)
-                .then(endpointOfSocketInfo)
+                .then(endpointOfSocketInfo);
         this.state_ = Connection.State.CONNECTING;
         this.onceConnected
             .then(() => {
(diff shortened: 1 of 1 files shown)


bemasc:
:+1:

iislucas:
I've a couple of things I'm not sure about here... 

`this.onceConnected` will now fulfill when it previously rejected. I think we it should still fail the same way; and we should add a test for this failure. 

Also, I think that if onceConnected does fail, then onceClosed should fulfill with `SocketCloseKind.NEVER_CONNECTED`

Sound sensible?

quantsword:
The catch is outside this.onceConnected promise. If it is previously rejected, it will still be rejected. 

I change the reason for close to be "SocketCloseKind.NEVER_CONNECTED"

src/tcp/tcp.ts
```diff
@@ -295,6 +295,14 @@ module Tcp {
               if(this.state_ !== Connection.State.CLOSED) {
                 this.state_ = Connection.State.CONNECTED;
               }
+            })
+            .catch((err) => {
+              log.error('Failed to connect to %1, err=%2',
+                        [JSON.stringify(connectionKind.endpoint),
+                         JSON.stringify(err)]);
+              this.dataToSocketQueue.stopHandling();
```
trevj, line 303: I don't think you need to do this since we only set a handler once `onceConnected` fulfills (a promise can't both fulfill and reject).

src/tcp/tcp.ts
```diff
@@ -295,6 +295,14 @@ module Tcp {
               if(this.state_ !== Connection.State.CLOSED) {
                 this.state_ = Connection.State.CONNECTED;
               }
+            })
+            .catch((err) => {
+              log.error('Failed to connect to %1, err=%2',
```
trevj, line 300: So, this isn't quite the "log and re-throw" anti-pattern but generally I've found it better to delegate logging of catchable/recoverable errors like this to the caller. Let's remove the logging statement here -- presumably our callers are already listening for rejection.

trevj:
Apologies for the delay in responding to this.

Thanks for the fix -- it looks right to me, even though as I said in the issue I can access the websites you mention via the proxy.

Regarding tests, we don't currently have unit tests for this file and I wouldn't want to block what looks like a good small fix on that.

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/172)

Reviewed files:
- src/tcp/tcp.ts @ r2

---

Looks like we're basically there - @trevj if you're happy give the thumbs-up. 

Only thing is that I can imagine connecting to bad addresses to be a normal part of operations - so probably console log wants to be a warning or info?

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/172)

---

:+1: once my comments have been addressed. Thanks again @quantsword!

---

iislucas:
These edits have basically been incorporated in various other pull requests now; thanks for these @quantsword ! Closing this pull request as it's not a bit stale w.r.t. branches and the work has been done on dev branch.

PR #173, opened 2015-01-05
Make start() reject if the connection is stopped before it starts.

bemasc:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/173)

### DIFF
--- a/src/socks-to-rtc/socks-to-rtc.spec.ts
+++ b/src/socks-to-rtc/socks-to-rtc.spec.ts
@@ -116,8 +116,8 @@ describe('SOCKS server', function() {
         server.stop).then(onceServerStopped).then(done);
   });
 
-  it('stop works before the PeerConnection connects', (done) => {
-    (<any>mockTcpServer.onceListening).and.returnValue(Promise.resolve(mockEndpoint));
+  it('stop works before the PeerConnection or TcpServer connects', (done) => {
+    (<any>mockTcpServer.onceListening).and.returnValue(noopPromise);
     // PeerConnection never connects.
     mockPeerConnection.onceConnected = noopPromise;
     // Neither TCP connection nor datachannel close "naturally".
@@ -126,7 +126,7 @@ describe('SOCKS server', function() {
     var onceStartFailed :Promise<void> = new Promise<void>((F, R) => {
       server.startInternal(mockTcpServer, mockPeerConnection).then(R, F);
     });
-    Promise.all([onceStartFailed, server.stop(), onceServerStopped()]).then(done);
+    Promise.all([onceStartFailed, server.stop()]).then(done);
   });
 });
 
--- a/src/socks-to-rtc/socks-to-rtc.spec.ts
+++ b/src/socks-to-rtc/socks-to-rtc.spec.ts
@@ -82,13 +82,28 @@ describe('SOCKS server', function() {
     server.startInternal(mockTcpServer, mockPeerConnection).catch(onceServerStopped).then(done);
   });
 
-  it('\'stopped\' fires on peerconnection termination', (done) => {
+  it('\'stopped\' fires, and start fails, on early peerconnection termination', (done) => {
     (<any>mockTcpServer.onceListening).and.returnValue(Promise.resolve(mockEndpoint));
-    (<any>mockTcpServer.onceShutdown).and.returnValue(Promise.resolve());
+    (<any>mockTcpServer.onceShutdown).and.returnValue(voidPromise);
     mockPeerConnection.onceConnected = voidPromise;
-    mockPeerConnection.onceDisconnected = <any>(Promise.resolve());
+    mockPeerConnection.onceDisconnected = voidPromise;
+
+    server.startInternal(mockTcpServer, mockPeerConnection).catch(onceServerStopped).then(done);
+  });
+
+  it('\'stopped\' fires on peerconnection termination', (done) => {
+    var terminate :() => void;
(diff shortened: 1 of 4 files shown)


bemasc:
@dborkan, this follows your test design.

I also had to make some changes to another test, and wound up splitting it into two cases.

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -82,13 +82,28 @@ describe('SOCKS server', function() {
     server.startInternal(mockTcpServer, mockPeerConnection).catch(onceServerStopped).then(done);
   });
 
+  it('\'stopped\' fires, and start fails, on early peerconnection termination', (done) => {
+    (<any>mockTcpServer.onceListening).and.returnValue(Promise.resolve(mockEndpoint));
+    (<any>mockTcpServer.onceShutdown).and.returnValue(voidPromise);
+    mockPeerConnection.onceConnected = voidPromise;
+    mockPeerConnection.onceDisconnected = voidPromise;
+
+    server.startInternal(mockTcpServer, mockPeerConnection).catch(onceServerStopped).then(done);
```
trevj, line 91: I'm trying to see how `startInternal` raises an error here but can't :-)

bemasc, line 91: Take a closer look at `rejectOnStopping` in socks-to-rtc.ts :-)

src/socks-to-rtc/socks-to-rtc.ts
```diff
@@ -163,7 +163,10 @@ module SocksToRtc {
       this.onceStopped_ = this.onceStopping_.then(this.stopResources_);
       this.onceStopped_.then(this.dispatchEvent_.bind(this, 'stopped'));
 
-      return onceReady;
+      var rejectOnStopping = new Promise((F, R) => {
```
trevj, line 166: Not related to this CL but I just realised the `onceStarted` field disappeared during the modularisation process. Not a big deal but sometimes it was nice to decouple it from the call to `start` -- particularly when we have >1 `start` methods.

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -100,6 +115,19 @@ describe('SOCKS server', function() {
     server.startInternal(mockTcpServer, mockPeerConnection).then(
         server.stop).then(onceServerStopped).then(done);
   });
+
+  it('stop works before the PeerConnection connects', (done) => {
+    (<any>mockTcpServer.onceListening).and.returnValue(Promise.resolve(mockEndpoint));
```
trevj: This is fine but I wonder if it wouldn't be clearer to have both TCP server and peerconnection use `noopPromise` here?

src/socks-to-rtc/socks-to-rtc.spec.ts
```diff
@@ -100,6 +115,19 @@ describe('SOCKS server', function() {
...
+    // PeerConnection never connects.
+    mockPeerConnection.onceConnected = noopPromise;
+    // Neither TCP connection nor datachannel close "naturally".
+    (<any>mockTcpServer.onceShutdown).and.returnValue(noopPromise);
+
+    var onceStartFailed :Promise<void> = new Promise<void>((F, R) => {
+      server.startInternal(mockTcpServer, mockPeerConnection).then(R, F);
+    });
+    Promise.all([onceStartFailed, server.stop(), onceServerStopped()]).then(done);
```
trevj: So it's a bit weird to check both `server.stop` and `onceServerStopped`. Not wrong but I would leave out `onceServerStopped` and I think it would still be fine.

trevj:
You're gonna hate me for saying this but at some point we should really do this for `RtcToNet`, too. I'd be happy to leave that for some future refactoring though, e.g. extracting a re-useable class for this highly intertwined stop/start behaviour.

It would be fine to leave as a TODO for now, though, I don't think the UI is waiting on that.

src/socks-to-rtc/socks-to-rtc.ts
bemasc, line 166: Freedom doesn't support fields, so it would have to be a Promise-getter method or a 'started' event.  Anyway, nothing was using it.  Currently we only really have one start method (`startInternal` is just for testing) so I think it's OK.

src/socks-to-rtc/socks-to-rtc.spec.ts
bemasc: Done.

src/socks-to-rtc/socks-to-rtc.spec.ts
bemasc: Done.

bemasc:
OK, I've added a TODO.

src/socks-to-rtc/socks-to-rtc.spec.ts
trevj, line 91: Oh...because we resolve the termination promises. Got it, thanks.

:+1:

src/socks-to-rtc/socks-to-rtc.ts
trevj, line 166: It's definitely fine for now and anyway my fault for missing in the original code review.

:+1:

trevj:
Should probably increase the NPM version, then :+1: .

bemasc:
OK, bumped minor version.

PR #174, opened 2015-01-06
Specify a minimum Chrome version for the sample apps.

bemasc:
The selected version is 41.0.2267.0, which is the first dev-channel
Chrome that has the fix for https://crbug.com/438699
(https://codereview.chromium.org/773243002), which caused
https://github.com/uProxy/uproxy/issues/615.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/174)


trevj:
:+1: 

Thanks!

PR #175, opened 2015-01-06
make shell script executable.

klazizpro:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/175)


trevj:
This is for the continuous build environment, right?

:+1:

PR #176, opened 2015-01-07
Code coverage

ryscheng:
Adding coverage reporting to turn-frontend, churn, socksToRtc, and rtcToNet

The coverage reports are generated as HTML files into
build/coverage

Currently here are the coverage numbers for these 4 directories
directory; statements, branches, functions, lines
churn/ 39.87%, 38%, 27.03%, 39.87%
turn-frontend/ 76.24%, 67.39%, 62.16%, 76.24%
rtcToNet/ 57.45%, 27.27%, 55.26%, 57.45%
socksToRtc/ 61.96%, 44.44%, 50.91%, 61.96%

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/176)

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -231,6 +231,13 @@ module.exports = (grunt) ->
         ])
         options:
           specs: 'build/turn-frontend/*.spec.js'
+          template: require('grunt-template-jasmine-istanbul')
+          templateOptions:
+            coverage: 'build/coverage/turn-frontend/coverage.json'
+            report:
+              type: 'html'
+              options:
+                dir: 'build/coverage/turn-frontend'
       # TODO: churn tests require peerconnection
       #       https://github.com/uProxy/uproxy/issues/430
       churn:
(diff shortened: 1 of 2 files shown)


ryscheng:
@trevj 
@mollyling
PTAL

trevj:
Thanks for taking the time to do this, @ryscheng!

I patched this in locally and it's fun and pretty revealing to browse the reports.

There's obviously some work we can do to template-acise this rule but I suspect the work we're doing with `require()` would help us first -- anyway, the hard part is figuring out how to wire the tools up, so this is a really great start.

One question: what would be needed to get the numbers onto, say, our Github page? Is that what coveralls would do (but we can't currently use it)?

ryscheng:
Yes. Coveralls does give you a "badge" that you can paste onto your GitHub README. As I mentioned in https://github.com/uProxy/uproxy/issues/491, it might be a bit difficult getting Coveralls to work with our TypeScript environment. 

The other option is to just publish these generated HTML files somewhere and link to them (e.g. in a dashboard). Since we don't have a dashboard, I wasn't sure where this should go.

We could also have a Grunt task that generates a little summary and automatically modifies our README file to reflect this. 

I suppose there's a few options. My favorite is to start working on a Dashboard somewhere and eventually publish to that.

trevj:
Oh yes, sorry, you did explicitly say that in the issue :-)

I agree that it might, In lieu of coveralls, it might be nice to include this on the dashboard.

@klazizpro, is this something that can be added to the new Jenkins dashboard?

ryscheng:
Are there any other concerns blocking a merge?

trevj:
Not from me. :+1:

ryscheng:
Cool thx

PR #177, opened 2015-01-07
Place generic type last as recommended by Typescript.

bemasc:
See https://github.com/uProxy/uproxy/issues/766

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/177)


iislucas:
:+1:

PR #178, opened 2015-01-07
add integration tests to grunt test

trevj:
So, we've had this integration test for the TCP wrapper since around the time we moved to Freedom 0.6. It was a little experiment and wasn't really reviewed (my bad: it got lumped in with some other changes). I'd like to get another pair of eyes on it now, and get it running as part of `grunt test`.

Then we can tackle the end-to-end SOCKS test:
https://github.com/uProxy/uproxy/issues/756

ONE THING: I've noticed this is currently flaky...tracking it down, but this test fails the first time the extension loads. If you reload the extension, it works fine. Just that first time it fails...I think if we could figure out why, it would explain why Simple SOCKS mysteriously fails to start occasionally:
https://github.com/uProxy/uproxy/issues/628

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/178)

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "freedom": "^0.6.11"
   },
   "scripts": {
-    "test": "grunt test",
+    "test": "grunt build jasmine",
     "prepublish": "grunt test",
     "benchmark": "node bin/run-benchmark.js -n 1024 -v"
   }
--- a/package.json
+++ b/package.json
@@ -44,7 +44,9 @@
     "utransformers": "~0.2.1",
     "wd": "~0.3.10",
     "wup": "^1.0.0",
-    "yargs": "^1.3.1"
+    "yargs": "^1.3.1",
+    "async": "^0.9.0",
+    "glob": "^4.3.2"
   },
   "peerDependencies": {
     "freedom": "^0.6.11"
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
   },
   "scripts": {
     "test": "grunt build jasmine",
-    "prepublish": "grunt test",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -50,8 +50,7 @@
     "freedom": "^0.6.11"
   },
   "scripts": {
-    "test": "grunt test",
-    "prepublish": "grunt test",
+    "test": "grunt build jasmine",
```
bemasc: What is the motivation for this change?

iislucas: I would expect this line to be 

```
 "test": "grunt test",
```

w.r.t. removal of `prepublish` I think that makes sense: it is not the right place for pre-submit tests and is deprecated.

trevj: So, I changed the test line to prevent Travis running the integration tests (it runs `npm test`). This is not the only way, e.g. I could add a new target like `grunt test_headless` but felt that was overkill.

bemasc: Is that because they're flaky, or because they're slow?

trevj: Neither: Chrome isn't available in the Travis environment so the integration tests can't be run. That's the only reason.

iislucas: OK, that makes sense. How about we make a `unit-tests` target and an `integration-tests` target, and each of those runs the relevant tests, and then it'll be super clear what's happening. :)

iislucas:
:+1:  from me (once you've picked your favorite solution for naming test build targets)

package.json
trevj: I factored out `unit_test` and `integration_test` from `grunt test`.

trevj:
Thanks for the review!

PR #179, opened 2015-01-09
Soycode copypastecrypto

agallant:
Ready to look at - you can now, if you choose, use the PGP api to sign/encrypt and verify/decrypt the copy paste payloads. Should be _mostly_ self explanatory - you get two new text areas, one with your public key and one for the public key of your friend. You exchange public keys first (so for now, more copy paste) - then when you generate the message it will automatically become a signed/encrypted PGP message, and when you paste your friend's signed/encrypted message it will automatically be decrypted.

Setup is automatic and right now just a silly test passphrase/user id (that's the same for all instances) - for this to be "real" we'd need to pull in a more meaningful id and such, and persist the keys across sessions etc. Still suitable for initial proof of concept I think.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/179)

### DIFF
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -160,7 +160,7 @@ freedom().on('friendKey', (newFriendKey:string) => {
 });
 
 freedom().on('signEncrypt', (message:string) => {
-  pgp.signEncrypt(str2ab(message), friendKey, false).then(//, true).then(
+  pgp.signEncrypt(str2ab(message), friendKey, true).then(
     function (cipherdata:ArrayBuffer) {
       pgp.armor(cipherdata).then(
         function (ciphertext:string) {
@@ -172,9 +172,9 @@ freedom().on('signEncrypt', (message:string) => {
 freedom().on('verifyDecrypt', (ciphertext:string) => {
   pgp.dearmor(ciphertext).then(
     function (cipherdata:ArrayBuffer) {
-      pgp.verifyDecrypt(cipherdata).then(//, friendKey).then(
+      pgp.verifyDecrypt(cipherdata, friendKey).then(
         function (result:VerifyDecryptResult) {
-          //freedom().emit('signed', result.signedBy[0] == 'Joe <joe@test.com>');
+          freedom().emit('signed', result.signedBy[0]);
           freedom().emit('plaintext', ab2str(result.data));
         });
     });
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -25,7 +25,6 @@ pgp.setup('super passphrase', 'Joe <joe@test.com>');
 pgp.exportKey().then(function (publicKey) {
   freedom().emit('publicKeyExport', publicKey);
 });
-//freedom().emit('signed', true);
 
 var rtcNetPcConfig :WebRtc.PeerConnectionConfig = {
   webrtcPcConfig: {
@@ -161,9 +160,24 @@ freedom().on('friendKey', (newFriendKey:string) => {
 });
 
 freedom().on('signEncrypt', (message:string) => {
-  pgp.signEncrypt(str2ab(message), friendKey).then(function (ciphertext) {
-    freedom().emit('ciphertext', ab2str(ciphertext));
-  });
+  pgp.signEncrypt(str2ab(message), friendKey, false).then(//, true).then(
+    function (cipherdata:ArrayBuffer) {
+      pgp.armor(cipherdata).then(
+        function (ciphertext:string) {
+          freedom().emit('ciphertext', ciphertext);
+        });
+    });
+});
+
+freedom().on('verifyDecrypt', (ciphertext:string) => {
+  pgp.dearmor(ciphertext).then(
+    function (cipherdata:ArrayBuffer) {
+      pgp.verifyDecrypt(cipherdata).then(//, friendKey).then(
+        function (result:VerifyDecryptResult) {
+          //freedom().emit('signed', result.signedBy[0] == 'Joe <joe@test.com>');
+          freedom().emit('plaintext', ab2str(result.data));
+        });
+    });
 });
 
 // Stops proxying.
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -24,6 +24,7 @@ pgp.setup('super passphrase', 'Joe <joe@test.com>');
 pgp.exportKey().then(function (publicKey) {
   freedom().emit('publicKeyExport', publicKey);
 });
+//freedom().emit('signed', true);
 
 var rtcNetPcConfig :WebRtc.PeerConnectionConfig = {
   webrtcPcConfig: {
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -141,11 +141,11 @@ freedom().on('handleSignalMessage', (signal:WebRtc.SignallingMessage) => {
 });
 
 // Crypto helper functions
-function ab2str(buf:ArrayBuffer) {
+function ab2str(buf:ArrayBuffer) :string {
   return String.fromCharCode.apply(null, new Uint16Array(buf));
 }
 
-function str2ab(str:string) {
+function str2ab(str:string) :ArrayBuffer {
   var buf = new ArrayBuffer(str.length * 2);  // 2 bytes for each char
   var bufView = new Uint16Array(buf);
   for (var i = 0, strLen = str.length; i < strLen; i++) {
@@ -161,20 +161,20 @@ freedom().on('friendKey', (newFriendKey:string) => {
 
 freedom().on('signEncrypt', (message:string) => {
   pgp.signEncrypt(str2ab(message), friendKey, true)
-    .then(function (cipherdata:ArrayBuffer) {
+    .then((cipherdata:ArrayBuffer) => {
       return pgp.armor(cipherdata);
     })
-    .then(function (ciphertext:string) {
+    .then((ciphertext:string) => {
       freedom().emit('ciphertext', ciphertext);
     });
 });
 
 freedom().on('verifyDecrypt', (ciphertext:string) => {
   pgp.dearmor(ciphertext)
-    .then(function (cipherdata:ArrayBuffer) {
+    .then((cipherdata:ArrayBuffer) => {
       return pgp.verifyDecrypt(cipherdata, friendKey);
     })
-    .then(function (result:VerifyDecryptResult) {
+    .then((result:VerifyDecryptResult) => {
       freedom().emit('signed', result.signedBy[0]);
       freedom().emit('plaintext', ab2str(result.data));
     });
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -22,7 +22,7 @@ var log :Logging.Log = new Logging.Log('copypaste-socks');
 var pgp :PgpProvider = freedom.pgp();
 var friendKey :string;
 // TODO interactive setup w/real passphrase
-pgp.setup('super passphrase', 'Joe <joe@test.com>');
+pgp.setup('', 'uProxy user <noreply@uproxy.org>');
 pgp.exportKey().then((publicKey:string) => {
   freedom().emit('publicKeyExport', publicKey);
 });
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -3,6 +3,7 @@
 
 /// <reference path='../../third_party/typings/es6-promise/es6-promise.d.ts' />
 /// <reference path='../../freedom/typings/pgp.d.ts' />
+/// <reference path='../../arraybuffers/arraybuffers.d.ts' />
 /// <reference path='../../logging/logging.d.ts' />
 /// <reference path='../../freedom/typings/freedom.d.ts' />
 /// <reference path='../../networking-typings/communications.d.ts' />
@@ -140,24 +141,13 @@ freedom().on('handleSignalMessage', (signal:WebRtc.SignallingMessage) => {
   }
 });
 
-// Helper function to convert strings to array buffers
-function str2ab(str:string) :ArrayBuffer {
-  // TODO: test more thoroughly w/unicode, standardize function across codebase
-  var buf = new ArrayBuffer(str.length * 2);  // 2 bytes for each char
-  var bufView = new Uint16Array(buf);
-  for (var i = 0, strLen = str.length; i < strLen; i++) {
-    bufView[i] = str.charCodeAt(i);
-  }
-  return buf;
-}
-
 // Crypto request messages
 freedom().on('friendKey', (newFriendKey:string) => {
   friendKey = newFriendKey;
 });
 
 freedom().on('signEncrypt', (message:string) => {
-  pgp.signEncrypt(str2ab(message), friendKey)
+  pgp.signEncrypt(ArrayBuffers.stringToArrayBuffer(message), friendKey)
     .then((cipherdata:ArrayBuffer) => {
       return pgp.armor(cipherdata);
     })
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -20,6 +20,11 @@ Logging.setConsoleFilter([
 
 var log :Logging.Log = new Logging.Log('copypaste-socks');
 var pgp :PgpProvider = freedom.pgp();
+// TODO interactive setup w/real passphrase
+pgp.setup('super passphrase', 'Joe <joe@test.com>');
+pgp.exportKey().then(function (publicKey) {
+  freedom().emit('publicKeyExport', publicKey);
+});
 
 var rtcNetPcConfig :WebRtc.PeerConnectionConfig = {
   webrtcPcConfig: {
@@ -52,12 +57,6 @@ var socksRtc:SocksToRtc.SocksToRtc;
 var rtcNet:RtcToNet.RtcToNet;
 
 freedom().on('start', () => {
-  // TODO interactive setup w/real passphrase
-  pgp.setup('super passphrase', 'Joe <joe@test.com>');
-  pgp.exportKey().then(function (publicKey) {
-    freedom().emit('publicKeyExport', publicKey);
-  });
-
   var localhostEndpoint:Net.Endpoint = { address: '127.0.0.1', port: 9999 };
   socksRtc = new SocksToRtc.SocksToRtc(
       localhostEndpoint,
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -1,6 +1,8 @@
 /// <reference path='../../rtc-to-net/rtc-to-net.d.ts' />
 /// <reference path='../../socks-to-rtc/socks-to-rtc.d.ts' />
 
+/// <reference path='../..//third_party/typings/es6-promise/es6-promise.d.ts' />
+/// <reference path='../../freedom/typings/pgp.d.ts' />
 /// <reference path='../../logging/logging.d.ts' />
 /// <reference path='../../freedom/typings/freedom.d.ts' />
 /// <reference path='../../networking-typings/communications.d.ts' />
@@ -17,6 +19,7 @@ Logging.setConsoleFilter([
     'RtcToNet:I']);
 
 var log :Logging.Log = new Logging.Log('copypaste-socks');
+var pgp :PgpProvider = freedom.pgp();
 
 var rtcNetPcConfig :WebRtc.PeerConnectionConfig = {
   webrtcPcConfig: {
@@ -49,6 +52,12 @@ var socksRtc:SocksToRtc.SocksToRtc;
 var rtcNet:RtcToNet.RtcToNet;
 
 freedom().on('start', () => {
+  // TODO interactive setup w/real passphrase
+  pgp.setup('super passphrase', 'Joe <joe@test.com>');
+  pgp.exportKey().then(function (publicKey) {
+    freedom().emit('publicKeyExport', publicKey);
+  });
+
   var localhostEndpoint:Net.Endpoint = { address: '127.0.0.1', port: 9999 };
   socksRtc = new SocksToRtc.SocksToRtc(
       localhostEndpoint,
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -22,7 +22,7 @@ var pgp :PgpProvider = freedom.pgp();
 var friendKey :string;
 // TODO interactive setup w/real passphrase
 pgp.setup('super passphrase', 'Joe <joe@test.com>');
-pgp.exportKey().then(function (publicKey) {
+pgp.exportKey().then((publicKey:string) => {
   freedom().emit('publicKeyExport', publicKey);
 });
 
@@ -140,12 +140,9 @@ freedom().on('handleSignalMessage', (signal:WebRtc.SignallingMessage) => {
   }
 });
 
-// Crypto helper functions
-function ab2str(buf:ArrayBuffer) :string {
-  return String.fromCharCode.apply(null, new Uint16Array(buf));
-}
-
+// Helper function to convert strings to array buffers
 function str2ab(str:string) :ArrayBuffer {
+  // TODO: test more thoroughly w/unicode, standardize function across codebase
   var buf = new ArrayBuffer(str.length * 2);  // 2 bytes for each char
   var bufView = new Uint16Array(buf);
   for (var i = 0, strLen = str.length; i < strLen; i++) {
@@ -160,7 +157,7 @@ freedom().on('friendKey', (newFriendKey:string) => {
 });
 
 freedom().on('signEncrypt', (message:string) => {
-  pgp.signEncrypt(str2ab(message), friendKey, true)
+  pgp.signEncrypt(str2ab(message), friendKey)
     .then((cipherdata:ArrayBuffer) => {
       return pgp.armor(cipherdata);
     })
@@ -175,8 +172,7 @@ freedom().on('verifyDecrypt', (ciphertext:string) => {
       return pgp.verifyDecrypt(cipherdata, friendKey);
     })
     .then((result:VerifyDecryptResult) => {
-      freedom().emit('signed', result.signedBy[0]);
-      freedom().emit('plaintext', ab2str(result.data));
+      freedom().emit('verifyDecryptResult', result);
     });
 });
 
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -1,7 +1,7 @@
 /// <reference path='../../rtc-to-net/rtc-to-net.d.ts' />
 /// <reference path='../../socks-to-rtc/socks-to-rtc.d.ts' />
 
-/// <reference path='../..//third_party/typings/es6-promise/es6-promise.d.ts' />
+/// <reference path='../../third_party/typings/es6-promise/es6-promise.d.ts' />
 /// <reference path='../../freedom/typings/pgp.d.ts' />
 /// <reference path='../../logging/logging.d.ts' />
 /// <reference path='../../freedom/typings/freedom.d.ts' />
@@ -146,9 +146,9 @@ function ab2str(buf:ArrayBuffer) {
 }
 
 function str2ab(str:string) {
-  var buf = new ArrayBuffer(str.length*2); // 2 bytes for each char
+  var buf = new ArrayBuffer(str.length * 2);  // 2 bytes for each char
   var bufView = new Uint16Array(buf);
-  for (var i=0, strLen=str.length; i<strLen; i++) {
+  for (var i = 0, strLen = str.length; i < strLen; i++) {
     bufView[i] = str.charCodeAt(i);
   }
   return buf;
@@ -160,23 +160,23 @@ freedom().on('friendKey', (newFriendKey:string) => {
 });
 
 freedom().on('signEncrypt', (message:string) => {
-  pgp.signEncrypt(str2ab(message), friendKey, true).then(
-    function (cipherdata:ArrayBuffer) {
-      pgp.armor(cipherdata).then(
-        function (ciphertext:string) {
-          freedom().emit('ciphertext', ciphertext);
-        });
+  pgp.signEncrypt(str2ab(message), friendKey, true)
+    .then(function (cipherdata:ArrayBuffer) {
+      return pgp.armor(cipherdata);
+    })
+    .then(function (ciphertext:string) {
+      freedom().emit('ciphertext', ciphertext);
     });
 });
 
 freedom().on('verifyDecrypt', (ciphertext:string) => {
-  pgp.dearmor(ciphertext).then(
-    function (cipherdata:ArrayBuffer) {
-      pgp.verifyDecrypt(cipherdata, friendKey).then(
-        function (result:VerifyDecryptResult) {
-          freedom().emit('signed', result.signedBy[0]);
-          freedom().emit('plaintext', ab2str(result.data));
-        });
+  pgp.dearmor(ciphertext)
+    .then(function (cipherdata:ArrayBuffer) {
+      return pgp.verifyDecrypt(cipherdata, friendKey);
+    })
+    .then(function (result:VerifyDecryptResult) {
+      freedom().emit('signed', result.signedBy[0]);
+      freedom().emit('plaintext', ab2str(result.data));
     });
 });
 
--- a/src/samples/copypaste-socks-chromeapp/freedom-module.ts
+++ b/src/samples/copypaste-socks-chromeapp/freedom-module.ts
@@ -19,6 +19,7 @@ Logging.setConsoleFilter([
 
 var log :Logging.Log = new Logging.Log('copypaste-socks');
 var pgp :PgpProvider = freedom.pgp();
+var friendKey :string;
 // TODO interactive setup w/real passphrase
 pgp.setup('super passphrase', 'Joe <joe@test.com>');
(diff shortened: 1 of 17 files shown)


src/samples/copypaste-socks-chromeapp/freedom-module.ts
```diff
@@ -131,6 +140,46 @@ freedom().on('handleSignalMessage', (signal:WebRtc.SignallingMessage) => {
   }
 });
 
+// Crypto helper functions
+function ab2str(buf:ArrayBuffer) {
+  return String.fromCharCode.apply(null, new Uint16Array(buf));
+}
+
+function str2ab(str:string) {
+  var buf = new ArrayBuffer(str.length*2); // 2 bytes for each char
```
bemasc: Spaces around *

src/samples/copypaste-socks-chromeapp/freedom-module.ts
```diff
@@ -131,6 +140,46 @@ freedom().on('handleSignalMessage', (signal:WebRtc.SignallingMessage) => {
...
+  var bufView = new Uint16Array(buf);
+  for (var i=0, strLen=str.length; i<strLen; i++) {
```
bemasc: Spaces around =

src/samples/copypaste-socks-chromeapp/freedom-module.ts
```diff
@@ -131,6 +140,46 @@ freedom().on('handleSignalMessage', (signal:WebRtc.SignallingMessage) => {
...
+
+// Crypto request messages
+freedom().on('friendKey', (newFriendKey:string) => {
+  friendKey = newFriendKey;
+});
+
+freedom().on('signEncrypt', (message:string) => {
+  pgp.signEncrypt(str2ab(message), friendKey, true).then(
+    function (cipherdata:ArrayBuffer) {
+      pgp.armor(cipherdata).then(
```
bemasc: The cool thing about Promises is that you can un-nest this kind of logic (here and below) by doing `return pgp.armor(cipherdata);'` and then chaining more `.then()` clauses off of it.

In fact, you could even do `.then(pgp.armor.bind(pgp)).then(...)`, if you find that clearer.

src/samples/copypaste-socks-chromeapp/freedom-module.ts
```diff
@@ -1,6 +1,8 @@
 /// <reference path='../../rtc-to-net/rtc-to-net.d.ts' />
 /// <reference path='../../socks-to-rtc/socks-to-rtc.d.ts' />
 
+/// <reference path='../..//third_party/typings/es6-promise/es6-promise.d.ts' />
```
bemasc: "//" ... typo?

src/samples/copypaste-socks-chromeapp/freedom-module.ts
bemasc: Two spaces before //

agallant:
Thanks, I believe I addressed the comments - and yeah I forgot about the better promise chaining style, though for consistency I decided to not do the bind. Thanks for taking a look!

src/samples/copypaste-socks-chromeapp/freedom-module.ts
```diff
@@ -131,6 +140,46 @@ freedom().on('handleSignalMessage', (signal:WebRtc.SignallingMessage) => {
...
+}
+
+// Crypto request messages
+freedom().on('friendKey', (newFriendKey:string) => {
+  friendKey = newFriendKey;
+});
+
+freedom().on('signEncrypt', (message:string) => {
+  pgp.signEncrypt(str2ab(message), friendKey, true)
+    .then(function (cipherdata:ArrayBuffer) {
```
bemasc: No space after function(.

bemasc: Actually, this probably shouldn't use `function` at all.  In general we prefer the fat-arrow (`=>`) functions, which are automatically bound.  This makes `this` a lot less confusing.

src/samples/copypaste-socks-chromeapp/freedom-module.ts
```diff
@@ -131,6 +140,46 @@ freedom().on('handleSignalMessage', (signal:WebRtc.SignallingMessage) => {
   }
 });
 
+// Crypto helper functions
+function ab2str(buf:ArrayBuffer) {
```
bemasc: `ab2str` and `str2ab` should declare a return type.

agallant:
Thanks, comments addressed and also added a README that hopefully addresses the questions from Lucas. So 6 copypastes is an exaggeration (that was back when I was also copypasting user ID around), it is 4 (2 public keys and 2 signaling messages). Will pointed out to me that it could be 3 (the getting side could bundle public key and their request message together), but that would require more tightly coupling crypto setup with giver/getter status - something that I think isn't actually desired in the "real world" use case (where we hopefully just have a better general story around key setup/exchange/storage).

Anyway, I tried to document that in the README, but further thoughts/feedback most definitely welcome.

willscott:
We could imagine that when keys aren't known, we default to the custom
3-step combined key-exchange / offer answer, and in subsequent interactions
with a known user you only need the 2 offer/answer steps since keying will
already be known.

--Will

agallant:
Sure that could make sense, but I guess I've been envisioning the public key exchange as more part of the "adding a buddy" process rather than the "I want to connect through my buddy right now" process. That said, since for most users the friends list will presumably be automatically populated, further streamlining may make sense. I'll ponder a bit more and look at the uproxy-proper code base for this sort of functionality, and if folks feel that this demo should work more like that let me know.

agallant:
Chatted a bit more with Will, and I think our consensus is that 3 steps makes sense, but it'd be best to do it in another demo (or uProxy itself). Because of the way this demo works (you can opt whether or not to use crypto) it's best to be able to do things separately for testing purposes, and bringing it down to 3 steps would multiply out the possible field behavior as well. In a separate crypto-only app it'd be simpler to support 3 steps.

agallant:
Revisiting this - are there any further changes or comments? Thanks!

iislucas:
@matthild Can you have a look? If not I'll make time and review tomorrow. Sorry for the delay!

matthild:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

---

<sup>**[Gruntfile.coffee, line 93 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoJdpu9C_UBw1vr9e3)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/Gruntfile.coffee#L93)):</sup>
Why is the source code in a directory called "build"?  The comment above refers to "compiled crypto lib .js", but what's going on here isn't entirely clear.  Can the source code move to less surprisingly named directory?  If not, can the comment be a bit more informative?

---

<sup>**[src/samples/copypaste-socks-chromeapp/crypto.html, line 20 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoKR6NrDZyQW_nARl2)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/crypto.html#L20)):</sup>
Please put the <bdi> element on the next line, indented, along with the rest of the content of the <p> element.  Likewise below.

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 25 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoKmAPK4llbco-UMZE)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L25)):</sup>
Most of the code uses =>, not 'function', for anonymous functions.  Probably best to be consistent.

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 149 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoN-ZF5VYpQi5SvEwl)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L149)):</sup>
Does this work for arbitrary Unicode text?  Apparently the 'length' property does not necessarily measure the number of characters in the string.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length

If yes, how about adding a comment to say so and briefly explain why?
If no, how about adding a TODO?

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 163 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoNuFWnL0YK31vfHwK)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L163)):</sup>
Please indicate what 'true' means here (e.g., by adding a comment or by passing a meaningfully named constant instead of the mysterious literal 'true').

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 178 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoPsnRTkOJJOo0r-cm)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L178)):</sup>
Please change "signed" to a more meaningful name, such as "signer" or "peerUserId" or some such.  (Or, better yet, eliminate it completely, as proposed below.)

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 179 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoSE-nhJLgSiRmlRAN)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L179)):</sup>
What do you think about consolidating these two into a single event?  For example:

freedom().emit('decryptedAndVerified',
               {'signer': result.signedBy[0], 'plaintext': ab2str(result.data)});

They go together conceptually, and reducing the number of event types simplifies the system.

---

<sup>**[src/samples/copypaste-socks-chromeapp/getter.ts, line 9 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoSlwlDKjTGbPiuB-y)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/getter.ts#L9)):</sup>
Parts of the code refer to encryption only while other parts refer to encryption & signing.  Please be consistent.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 22 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoQMqfUxStVEvMoexp)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L22)):</sup>
Decrypted what?  Can this field have a more informative name?  Likewise for the 'signed' field below.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 32 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoQgl0DTYV9lOQGuW_)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L32)):</sup>
I think it would be a bit clearer to put all the crypto-related fields together.  Is there any reason not to?

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 113 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoT75cvuOIlKXjDx9v)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L113)):</sup>
What do you think about combining these two into a single event (e.g., 'receivedArmoredMessage')?

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 130 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoTKnZNATwRmxHVzty)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L130)):</sup>
Combine into a single event (e.g., 'armorMessage' or whatnot)?

---

<sup>**[src/samples/copypaste-socks-chromeapp/README.md, line 9 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoRG5MzQD4oFBrlAGA)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/README.md#L9)):</sup>
- They will be encrypted and also signed, right?
- It's a bit confusing to say "with a randomly generated keypair" because there are actually TWO randomly generated keypairs involved, right?  (Each peer has one.)

---

agallant:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

---

<sup>**[Gruntfile.coffee, line 93 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoJdpu9C_UBw1vr9e3)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/Gruntfile.coffee#L93)):</sup>
This is due to how the Gruntfile for freedom-pgp-e2e is structured: https://github.com/freedomjs/freedom-pgp-e2e/blob/master/Gruntfile.js

It's fairly similar to the other Gruntfiles/repos we use, though the more complicated ones put the final result in a dist/ folder instead (and build/ is more of an intermediary). I could rename this build/ to dist/ if that seems clearer in this context, let me know if that addresses your concern.

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 25 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoKmAPK4llbco-UMZE)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L25)):</sup>
Done.

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 149 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoN-ZF5VYpQi5SvEwl)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L149)):</sup>
Done - went with TODO, and also note that this function (or something very similar) is used in various other places throughout the greater codebase - so probably at some point getting a standard interface and implementation of translating between strings, arraybuffers, and so forth would be good (which is beyond the scope of this one file, hence the TODO).

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 163 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoNuFWnL0YK31vfHwK)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L163)):</sup>
I'll just go ahead and remove the true as that's the default parameter anyway, and it's a sensible default (it just means to sign as well as encrypt - the API is specified in the .json and is similar to the underlying e2e crypto library) and probably what is expected by reading something named signEncrypt.

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 178 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoPsnRTkOJJOo0r-cm)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L178)):</sup>
See next comment.

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 179 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoSE-nhJLgSiRmlRAN)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L179)):</sup>
Done - however, it required moving the ab2str helper function to popup.ts, to use the VerifyDecryptResult type rather than creating a new type. See: https://github.com/uProxy/uproxy-lib/blob/dev/src/freedom/typings/pgp.d.ts#L10

---

<sup>**[src/samples/copypaste-socks-chromeapp/getter.ts, line 9 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoSlwlDKjTGbPiuB-y)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/getter.ts#L9)):</sup>
Done - personally I consider "verifyDecryptInboundMessage" to be a bit of a mouthful, but not strongly enough to object if you think it clarifies things.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 22 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoQMqfUxStVEvMoexp)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L22)):</sup>
Renaming to prefix with "input" - though I'd say it was actually fairly clear from the context and purpose of this application.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 32 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoQgl0DTYV9lOQGuW_)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L32)):</sup>
Done.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 113 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoT75cvuOIlKXjDx9v)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L113)):</sup>
Doing so would be a tradeoff, in that we'd need to add a new type/interface (the previous VerifyDecrypt combination could leverage an existing type at least). We also may end up wanting a separate publicKey message type regardless, though if we can definitely factor those out I'd consider combining. Let me know your thoughts re: adding another type.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 130 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoTKnZNATwRmxHVzty)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L130)):</sup>
Again, would lead to more types - so we're weighing a proliferation of messages against a proliferation of types. In this case I'd rather go with the simple atomic messages and basic types, but if there's an actual cost/overhead to the messages besides the extra line of code then I'd consider a refactor.

---

<sup>**[src/samples/copypaste-socks-chromeapp/README.md, line 9 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoRG5MzQD4oFBrlAGA)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/README.md#L9)):</sup>
Done.

---

agallant:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

---

Replying to this so Reviewable stops making it red.

---

<sup>**[src/samples/copypaste-socks-chromeapp/crypto.html, line 20 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoKR6NrDZyQW_nARl2)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/crypto.html#L20)):</sup>
Done.

---

agallant:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

---

<sup>**[Gruntfile.coffee, line 93 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoJdpu9C_UBw1vr9e3)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/Gruntfile.coffee#L93)):</sup>
Thinking about this a bit more, perhaps the right approach is to only npm publish within the build/ directory, but I'm not sure if that's the norm or if it's beneficial to publish the whole repo (I do think it is more typical at least). Will think about it more, thoughts are welcome - thanks.

---

matthild:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

---

<sup>**[Gruntfile.coffee, line 93 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoJdpu9C_UBw1vr9e3)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/Gruntfile.coffee#L93)):</sup>
I'm not so familiar with Gruntifles, actually, and just found it surprising that the source code was in a directory called "build".  If you think it's clear as-is, that's fine with me.  If there's something unusual about freedom-pgp-e2e's directory structure, perhaps making the comment a bit more informative would be helpful.

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 176 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoSE-nhJLgSiRmlRAN)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L176)):</sup>
If the other two candidate locations for event consolidation (below) remain as pairs of separate events, do you think this event here should remain consolidated or revert to being a pair for consistency?

---

<sup>**[src/samples/copypaste-socks-chromeapp/getter.ts, line 9 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoSlwlDKjTGbPiuB-y)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/getter.ts#L9)):</sup>
It's certainly clearer, but certainly longer too.  If you find it too long, perhaps some alternative (but still consistent) vocabulary would be helpful -- e.g., perhaps "protect" and "unprotect" could mean "sign & encrypt" and "decrypt & verify", respectively.  (Personally I think either way is fine -- just a thought related to your point about length.)

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 22 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoQMqfUxStVEvMoexp)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L22)):</sup>
Hm, you are right. If you prefer the old names, that's fine with me.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 118 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoT75cvuOIlKXjDx9v)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L118)):</sup>
Whatever you think best is fine with me.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 135 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoTKnZNATwRmxHVzty)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L135)):</sup>
As above, whatever you think best is fine with me.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

---

<sup>**[Gruntfile.coffee, line 93 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoJdpu9C_UBw1vr9e3)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/Gruntfile.coffee#L93)):</sup>
There's some work to rationalize the build system; we were using sym-links to work around bugs/missing things in grunt-ts. I've since found a better way to doing this, but that's a separate pull request. So lets let that get fixed elsewhere. 

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 146 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoN-ZF5VYpQi5SvEwl)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L146)):</sup>
This is already done in the uproxy arraybuffers util library, please use the version from there. 

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 117 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-Jh-vwivlNhujyud7Ipb)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5faa1f86155c1dca4944617c6c0e241c4abcadb7/src/samples/copypaste-socks-chromeapp/popup.ts#L117)):</sup>
Don't use `any` type unless you need to. It doens't tell the reader of the code anything helpful, and it sets you up for errors later. 

---

agallant:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 146 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoN-ZF5VYpQi5SvEwl)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L146)):</sup>
Done.

---

<sup>**[src/samples/copypaste-socks-chromeapp/freedom-module.ts, line 176 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoSE-nhJLgSiRmlRAN)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/freedom-module.ts#L176)):</sup>
I'm okay either way - I was fine converting this one as we already needed that type and had it defined.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 22 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JgoQMqfUxStVEvMoexp)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d2c7e8f1603b4d572c0e009f6ca0594d2a6b43b8/src/samples/copypaste-socks-chromeapp/popup.ts#L22)):</sup>
I'm okay with these, they're consistent with the other input-prefixed ones.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 117 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-Jh-vwivlNhujyud7Ipb)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5faa1f86155c1dca4944617c6c0e241c4abcadb7/src/samples/copypaste-socks-chromeapp/popup.ts#L117)):</sup>
Agreed, that type was there and is in a few other places so I wasn't sure what it really was myself. Per a bit of logging it is evidently:
EventInterface
-emit: function () { [native code] }
-off: function () { [native code] }
-on: function () { [native code] }
-once: function () { [native code] }
-**proto**: EventInterface

On first glance I didn't find a type like this specified elsewhere in the uProxy codebase, but it looks elemental enough that I feel like I may be missing it. Are you aware of such a type binding, or if I do define it where would you suggest I do so? Thanks!

---

matthild:
:+1: from me, but please wait for @iislucas to follow up regarding the 'any' type.  Thanks!

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

Reviewed files:
- Gruntfile.coffee @ r1
- package.json @ r1
- src/samples/copypaste-socks-chromeapp/crypto.html @ r5
- src/samples/copypaste-socks-chromeapp/crypto.ts @ r1
- src/samples/copypaste-socks-chromeapp/freedom-module.json @ r1
- src/samples/copypaste-socks-chromeapp/freedom-module.ts @ r7
- src/samples/copypaste-socks-chromeapp/getter.html @ r1
- src/samples/copypaste-socks-chromeapp/getter.ts @ r4
- src/samples/copypaste-socks-chromeapp/giver.html @ r1
- src/samples/copypaste-socks-chromeapp/giver.ts @ r6
- src/samples/copypaste-socks-chromeapp/locales/en/messages.json @ r1
- src/samples/copypaste-socks-chromeapp/popup.ts @ r7
- src/samples/copypaste-socks-chromeapp/README.md @ r4
- src/samples/copypaste-socks-chromeapp/root.html @ r1
- src/samples/copypaste-socks-chromeapp/root.ts @ r1

---

Very close now! one suggested edit to way messages are handled, and one suggestion for the type cleanup.

---

<sup>**[src/samples/copypaste-socks-chromeapp/locales/en/messages.json, line 17 [r7]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JhD61LUr-SM5b5mBBzZ)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/b346933842a7493e82a9c90f44dee13cc3c3f422/src/samples/copypaste-socks-chromeapp/locales/en/messages.json#L17)):</sup>
Do you want a string for when there is not yet a signaling message/it is empty as distinct from saying it doesn't have a valid signature? I'd be inlined to use the phrase `incorrectlySigned` for this also... so it's more consistent with the string that's quite scary sounding. :) 

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 113 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-Jh-vwivlNhujyud7Ipb)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5faa1f86155c1dca4944617c6c0e241c4abcadb7/src/samples/copypaste-socks-chromeapp/popup.ts#L113)):</sup>
Hrm... OnAndEmit (here: https://github.com/uProxy/uproxy-lib/blob/dev/src/freedom/typings/freedom.d.ts#L15 ) was intended to capture this, but it had forgotten about `once` and `on` and `off` and `proto`.  I think you only use those, so go with that for now, and I'll do a followup pull request to uProxy for an OnEmitStub which includes off and on and proto (can you email me what those do, and I'll then include docs in the freedom typescript definition too)

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

---

by the way: I tried it today and I think it is awesome. :) 

---

agallant:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

---

<sup>**[src/samples/copypaste-socks-chromeapp/locales/en/messages.json, line 17 [r7]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-JhD61LUr-SM5b5mBBzZ)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/b346933842a7493e82a9c90f44dee13cc3c3f422/src/samples/copypaste-socks-chromeapp/locales/en/messages.json#L17)):</sup>
Done - and for now I'm just hiding any message about the signature (valid or invalid) unless inputIsWellFormed is true, as it's only really relevant then (when there's a possibility of the user connecting). There's a TODO about letting the user know input is malformed but I'll punt on that as that's perhaps a separate UX/design question anyway.

---

<sup>**[src/samples/copypaste-socks-chromeapp/popup.ts, line 113 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179#-Jh-vwivlNhujyud7Ipb)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5faa1f86155c1dca4944617c6c0e241c4abcadb7/src/samples/copypaste-socks-chromeapp/popup.ts#L113)):</sup>
Done.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/179)

Reviewed files:
- Gruntfile.coffee @ r8
- src/samples/copypaste-socks-chromeapp/connection-state.ts @ r8
- src/samples/copypaste-socks-chromeapp/crypto.html @ r8
- src/samples/copypaste-socks-chromeapp/freedom-module.ts @ r8
- src/samples/copypaste-socks-chromeapp/getter.ts @ r8
- src/samples/copypaste-socks-chromeapp/giver.ts @ r8
- src/samples/copypaste-socks-chromeapp/locales/en/messages.json @ r8
- src/samples/copypaste-socks-chromeapp/popup.html @ r8
- src/samples/copypaste-socks-chromeapp/popup.ts @ r8

---

LGTM! :+1:

---

PR #181, opened 2015-01-14
Integration tests for SocksToRtc and RtcToNet.

bemasc:
This resolves https://github.com/uProxy/uproxy/issues/756

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/181)


src/integration-tests/socks-echo/integration.ts
```diff
@@ -0,0 +1,168 @@
...
+                }
+              });
+            });
+          });
+    } catch (e) {
+      return Promise.reject(e.message + ' ' + e.stack);
+    }
+  }
+
+  public parallelEchoTest = (contents:ArrayBuffer[]) : Promise<ArrayBuffer[]> => {
```
bemasc: Something funny is going on here... this should be serialEchoTest, and the test should fail!

bemasc: Resolved.  (parallelEchoTest was defined twice here, but there was a matching typo in the jasmine tests that call parallelEchoTest in both the "parallel" and "serial" tests.  I've fixed both sides and confirmed that the tests still pass.)

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181)

---

Oh, this is great. Extremely happy to see this coming. I just have a few comments on details, nothing major. Great stuff.

---

<sup>**[Gruntfile.coffee, line 189 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfednHd088FzE5oEq3q)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/Gruntfile.coffee#L189)):</sup>
So, I think it would be fine to name this directory `integration`. That would be a bit shorter and I see it's what they've done in the freedom-for-\* repositories, too. Not a big deal and I know it's a pain, but I wonder what you think?

---

<sup>**[src/integration-tests/socks-echo/integration.json, line 21 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfeoNFQNCFsnfM1ZX3g)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.json#L21)):</sup>
So having a custom API obviously makes the tests easier to read and understand which is great. I opted against it for the TCP tests because I figured it was just too much hassle to get right....if that wasn't the case for you, great. I just want as few barriers to writing tests as possible!

---

<sup>**[src/integration-tests/socks-echo/integration.ts, line 65 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfemIeTGPB0vUE7KYcT)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.ts#L65)):</sup>
How about a little more detail here, e.g. `socks server returned unexpected AUTH value`?

---

<sup>**[src/integration-tests/socks-echo/integration.ts, line 82 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfemQdAZzN1Bo12eTbp)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.ts#L82)):</sup>
Again, a teeny bit more detail here might help, e.g. `SOCKS server connected to wrong address`.

---

<sup>**[src/integration-tests/socks-echo/integration.ts, line 104 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-Jfem6aF7r8Zwdx-Yb17)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.ts#L104)):</sup>
Are you doing this so that the message and stacktrace can traverse the Freedom module boundary?

---

<sup>**[src/integration-tests/socks-echo/integration.ts, line 153 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-Jfeks_lwHByReSlqDY6)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.ts#L153)):</sup>
Nice async `for` loop pattern :-)

---

<sup>**[src/integration-tests/socks-echo/socks-echo.integration.spec.ts, line 9 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfefDPjUP7hFfjeptRf)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/socks-echo.integration.spec.ts#L9)):</sup>
You can use the functions in `ArrayBuffers` here.

---

<sup>**[src/integration-tests/socks-echo/socks-echo.integration.spec.ts, line 29 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfefdXS_m9-XtavkBFP)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/socks-echo.integration.spec.ts#L29)):</sup>
Are you sure this does what you think? I played around with `toEqual` and `Uint8Array` while working on TURN support and concluded it just doesn't work:
https://github.com/uProxy/uproxy-networking/blob/dev/src/turn-frontend/messages.spec.ts#L364

Let's double check this actually works!

---

<sup>**[src/integration-tests/socks-echo/socks-echo.integration.spec.ts, line 31 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfegIr5jtBKyjQQwMbr)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/socks-echo.integration.spec.ts#L31)):</sup>
I don't think you need this: if the previous block raises an exception then the test will timeout and fail.

---

<sup>**[src/integration-tests/socks-echo/socks-echo.integration.spec.ts, line 47 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-Jfel6TALoWY4hKinixO)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/socks-echo.integration.spec.ts#L47)):</sup>
So, it's definitely good to test concurrent requests -- or as close to it as we can manage in JS.

Something we're not checking here, however, is that each connection receives its own response -- do you agree that's a pretty important part of handling concurrent requests? I think it's probably not super-hard to re-write `parallelEchoTest` to do this...what do you think?

---

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181)

---

<sup>**[Gruntfile.coffee, line 189 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfednHd088FzE5oEq3q)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/Gruntfile.coffee#L189)):</sup>
Done.

---

<sup>**[src/integration-tests/socks-echo/integration.json, line 21 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfeoNFQNCFsnfM1ZX3g)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.json#L21)):</sup>
It was definitely trickier than I expected, but it was a good learning experience, and hopefully will be easy to cargo-cult for future integration tests.

---

<sup>**[src/integration-tests/socks-echo/integration.ts, line 0 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-Jfeks_lwHByReSlqDY6)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.ts#L0)):</sup>
Thanks :-)  It's now moved into socks-echo.integration.spec.ts, and possibly improved a little.

---

<sup>**[src/integration-tests/socks-echo/integration.ts, line 0 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-Jfem6aF7r8Zwdx-Yb17)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.ts#L0)):</sup>
Yes.  This makes it much, much easier to debug failures.  Otherwise, the promise just never resolves, and the test fails with a timeout.

I wish that Freedom would do this automatically!

---

<sup>**[src/integration-tests/socks-echo/integration.ts, line 65 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfemIeTGPB0vUE7KYcT)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.ts#L65)):</sup>
Done.

---

<sup>**[src/integration-tests/socks-echo/integration.ts, line 82 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfemQdAZzN1Bo12eTbp)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.ts#L82)):</sup>
Done.

---

<sup>**[src/integration-tests/socks-echo/socks-echo.integration.spec.ts, line 0 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfefDPjUP7hFfjeptRf)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/socks-echo.integration.spec.ts#L0)):</sup>
Done.

---

<sup>**[src/integration-tests/socks-echo/socks-echo.integration.spec.ts, line 0 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfefdXS_m9-XtavkBFP)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/socks-echo.integration.spec.ts#L0)):</sup>
I'm pretty sure `toEqual` works with `Uint8Array`, because I saw it fail during development, but for clarity I've switched all these checks to use `ArrayBuffers.byteEquality`.

---

<sup>**[src/integration-tests/socks-echo/socks-echo.integration.spec.ts, line 0 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfegIr5jtBKyjQQwMbr)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/socks-echo.integration.spec.ts#L0)):</sup>
Yes, but this way the contents of the exception is actually shown as the cause of the failure, which makes it much easier to debug failures.

I wish jasmine did this automatically!

---

<sup>**[src/integration-tests/socks-echo/socks-echo.integration.spec.ts, line 0 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-Jfel6TALoWY4hKinixO)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/socks-echo.integration.spec.ts#L0)):</sup>
Good point!  After further consideration I realized that this test was not really concurrent anyway.  I've restructured the tests to enable truly concurrent tests, and to ensure that the right input is compared to its corresponding output.

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181)

---

I love it. Creating multiple echo servers, in particular, is super cool.

:+1:

---

<sup>**[Gruntfile.coffee, line 325 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfjD8QNVhXT-8BS7QbJ)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6cdf7c64cc9032cca2fbf856fc8e5108ff098228/Gruntfile.coffee#L325)):</sup>
Very helpful comment.

---

<sup>**[src/integration-tests/socks-echo/integration.ts, line 0 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-Jfem6aF7r8Zwdx-Yb17)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.ts#L0)):</sup>
Cool, good to know :-)

---

<sup>**[src/integration-tests/socks-echo/integration.ts, line 0 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfemIeTGPB0vUE7KYcT)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/integration.ts#L0)):</sup>
Nice. This is one time we really miss having `expect`!

---

<sup>**[src/integration-tests/socks-echo/socks-echo.integration.spec.ts, line 0 [r5]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfefdXS_m9-XtavkBFP)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ce4563c5e067f0ff8c203114e293e1090ebb1df5/src/integration-tests/socks-echo/socks-echo.integration.spec.ts#L0)):</sup>
Nice, I didn't even know that existed.

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 50 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/181#-JfjHd8mM_sewII2JN9q)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6cdf7c64cc9032cca2fbf856fc8e5108ff098228/src/integration/socks-echo/socks-echo.integration.spec.ts#L50)):</sup>
It seems like this is the only test that calls `echo` with an `ArrayBuffer`. You might think about making it just accept and return a single Object and making this test do its own batching (every other test has to extract the zero-th element from the return value, which is maybe not super-pretty).

It's fine as-is though, I just want to suggest it.

---

PR #182, opened 2015-01-20
Added keepalive to simple-socks-chromeapp

mollyling:
Referencing https://github.com/uProxy/uproxy/issues/742

This keeps the app from sleeping.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/182)

### DIFF
--- a/src/samples/simple-socks-chromeapp/background.ts
+++ b/src/samples/simple-socks-chromeapp/background.ts
@@ -5,10 +5,13 @@ script.src = 'lib/freedom/freedom-for-chrome.js';
 document.head.appendChild(script);
 
 script.onload = () => {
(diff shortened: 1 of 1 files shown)


trevj:
Unfortunately, I can imagine the app being suspended in <5 seconds!

Let's:
1. Make that 1 second (will that kill the VM? I doubt it).
2. Add a quick comment for why we do this crazy thing.

Then, :+1: . Thanks!

mollyling:
As long as their's a pending alarm, it should keep the thing open.  We could knock it up to 5 minutes -- and that'd be more efficient, as it runs the JS less often while keeping the app alive.

trevj:
Good point, thanks :-) :+1: for whatever timeout you think is best.

PR #183, opened 2015-01-20
Fixed median calc, updated timeout, and parameterized num timeouts.

mollyling:
Small changes to make the benchmark tool better.  Specifically:
- I turned length-checking back on.  I after having diagnosed the length issue in https://github.com/uProxy/uproxy/issues/770
- Upped the timeout for requests to 30 seconds from 2 sec.  There were requests on localhost taking a surprisingly long amount of time, but let's make sure that we know if they eventually come back.
- Fixed median latency calculation. 

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/183)

### DIFF
--- a/bin/run-benchmark.js
+++ b/bin/run-benchmark.js
@@ -10,6 +10,8 @@ var argv = require('yargs')
     .alias('c', 'concurrency')
     .default('speed', 1)
     .alias('s', 'speed')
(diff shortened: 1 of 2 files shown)


trevj:
:+1:

mollyling:
Thanks!

PR #184, opened 2015-01-20
Clarify socks integration test by adding a single-echo helper function.

bemasc:
Addresses https://github.com/uProxy/uproxy/issues/782

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/184)


trevj:
:+1:

PR #185, opened 2015-01-21
Fix local IP address filtering in RtcToNet.

bemasc:
Fix for https://github.com/uProxy/uproxy/issues/780

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/185)

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -336,7 +336,7 @@ module.exports = (grunt) ->
             {path: 'build/tcp/tcp.js', include: false}
             {path: 'build/integration/*/integration.*', include: false}
           ]
-          keepBrowser: false
+          keepBrowser: true
 
     clean: ['build/', 'dist/', '.tscache/']
 
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -336,7 +336,7 @@ module.exports = (grunt) ->
             {path: 'build/tcp/tcp.js', include: false}
             {path: 'build/integration/*/integration.*', include: false}
           ]
-          keepBrowser: true
+          keepBrowser: false
 
     clean: ['build/', 'dist/', '.tscache/']
 
--- a/src/integration/socks-echo/integration.json
+++ b/src/integration/socks-echo/integration.json
@@ -21,14 +21,17 @@
   "default": "ProxyIntegrationTest",
   "api": {
     "ProxyIntegrationTest": {
+      "constructor": {
+        "value": ["boolean"]
+      },
       "startEchoServer": {
         "type": "method",
-        "value": ["string"],
-        "ret": []
+        "value": [],
+        "ret": "number"
       },
       "connect": {
         "type": "method",
-        "value": ["string"],
+        "value": ["number", "string"],
         "ret": "string"
       },
       "echo": {
--- a/src/integration/socks-echo/integration.ts
+++ b/src/integration/socks-echo/integration.ts
@@ -9,16 +9,18 @@ class ProxyIntegrationTest {
   private socksToRtc_ :SocksToRtc.SocksToRtc;
   private rtcToNet_ :RtcToNet.RtcToNet;
   private socksEndpoint_ : Promise<Net.Endpoint>;
-  private echoServers_ :{ [index:string]: Tcp.Server; } = {};
+  private echoServers_ :Tcp.Server[] = [];
   private connections_ :{ [index:string]: Tcp.Connection; } = {};
+  private localhost_ :string = '127.0.0.1';
 
-  constructor(private dispatchEvent_:(name:string, args:any) => void) {
-    this.socksEndpoint_ = this.startSocksPair_();
+  constructor(private dispatchEvent_:(name:string, args:any) => void,
+                                      denyLocalhost?:boolean) {
+    this.socksEndpoint_ = this.startSocksPair_(denyLocalhost);
   }
 
-  public startEchoServer = (name:string) : Promise<void> => {
+  public startEchoServer = () : Promise<number> => {
     var server = new Tcp.Server({
-      address: '127.0.0.1',
+      address: this.localhost_,
       port: 0
     });
 
@@ -29,13 +31,13 @@ class ProxyIntegrationTest {
     });
 
     // Discard endpoint info; we'll get it again later via .onceListening().
-    this.echoServers_[name] = server;
-    return server.listen().then((endpoint:Net.Endpoint) => {});
+    this.echoServers_.push(server);
+    return server.listen().then((endpoint:Net.Endpoint) => { return endpoint.port; });
   }
 
-  private startSocksPair_ = () : Promise<Net.Endpoint> => {
+  private startSocksPair_ = (denyLocalhost?:boolean) : Promise<Net.Endpoint> => {
     var socksToRtcEndpoint :Net.Endpoint = {
-      address: '127.0.0.1',
+      address: this.localhost_,
       port: 0
     };
     var socksToRtcPcConfig :WebRtc.PeerConnectionConfig = {
@@ -49,7 +51,7 @@ class ProxyIntegrationTest {
       initiateConnection: false
     };
     var rtcToNetProxyConfig :RtcToNet.ProxyConfig = {
-      allowNonUnicast: true  // Allow RtcToNet to contact the localhost server.
+      allowNonUnicast: !denyLocalhost  // Allow RtcToNet to contact the localhost server.
     };
 
     this.socksToRtc_ = new SocksToRtc.SocksToRtc();
@@ -64,7 +66,14 @@ class ProxyIntegrationTest {
     var connection = new Tcp.Connection({endpoint: socksEndpoint});
     var authRequest = Socks.composeAuthHandshakeBuffer([Socks.Auth.NOAUTH]);
     connection.send(authRequest);
-    return connection.receiveNext().then((buffer:ArrayBuffer) : Promise<ArrayBuffer> => {
+    var connected = new Promise<Tcp.ConnectionInfo>((F, R) => {
+      connection.onceConnected.then(F);
+      connection.onceClosed.then(R);
+    });
+    var firstBufferPromise :Promise<ArrayBuffer> = connection.receiveNext();
+    return connected.then((i:Tcp.ConnectionInfo) => {
+      return firstBufferPromise;
+    }).then((buffer:ArrayBuffer) : Promise<ArrayBuffer> => {
       var auth = Socks.interpretAuthResponse(buffer);
       if (auth != Socks.Auth.NOAUTH) {
         throw new Error('SOCKS server returned unexpected AUTH response.  ' +
@@ -72,40 +81,27 @@ class ProxyIntegrationTest {
       }
 
       var request :Socks.Request = {
-        version: Socks.VERSION5,
         command: Socks.Command.TCP_CONNECT,
-        destination: {
-          addressType: Socks.AddressType.IP_V4,
-          endpoint: webEndpoint,
-          addressByteLength: 7
-        }
+        endpoint: webEndpoint,
       };
       connection.send(Socks.composeRequestBuffer(request));
       return connection.receiveNext();
-    }).then((buffer:ArrayBuffer) : Tcp.Connection => {
-      var responseEndpoint = Socks.interpretRequestResponse(buffer);
-      if (responseEndpoint.address != webEndpoint.address) {
-        throw new Error('SOCKS server connected to wrong address.  ' +
-                        'Expected ' + webEndpoint.address +
-                        ' but got ' + responseEndpoint.address);
+    }).then((buffer:ArrayBuffer) : Promise<Tcp.Connection> => {
+      var reply = Socks.interpretReplyBuffer(buffer);
+      if (reply.replyField != Socks.Response.SUCCEEDED) {
+        return Promise.reject(reply);
       }
-      if (responseEndpoint.port != webEndpoint.port) {
-        throw new Error('SOCKS server connected to wrong port.  ' +
-                        'Expected ' + webEndpoint.port +
-                        ' but got ' + responseEndpoint.port);
-      }
-      return connection;
+      return Promise.resolve(connection);
     });
   }
 
-  public connect = (echoServerName:string) : Promise<string> => {
+  public connect = (port:number, address?:string) : Promise<string> => {
     try {
-      return Promise.all([
-        this.socksEndpoint_,
-        this.echoServers_[echoServerName].onceListening()
-      ]).then((endpoints:Net.Endpoint[]) : Promise<Tcp.Connection> => {
-        var socksEndpoint = endpoints[0];
-        var echoEndpoint = endpoints[1];
+      return this.socksEndpoint_.then((socksEndpoint:Net.Endpoint) : Promise<Tcp.Connection> => {
+        var echoEndpoint :Net.Endpoint = {
+          address: address || this.localhost_,
+          port: port
+        };
         return this.connectThroughSocks_(socksEndpoint, echoEndpoint);
       }).then((connection:Tcp.Connection) => {
         this.connections_[connection.connectionId] = connection;
--- a/src/integration/socks-echo/integration.ts
+++ b/src/integration/socks-echo/integration.ts
@@ -87,9 +87,9 @@ class ProxyIntegrationTest {
       connection.send(Socks.composeRequestBuffer(request));
       return connection.receiveNext();
     }).then((buffer:ArrayBuffer) : Promise<Tcp.Connection> => {
-      var reply = Socks.interpretReplyBuffer(buffer);
-      if (reply.replyField != Socks.Response.SUCCEEDED) {
-        return Promise.reject(reply);
+      var response = Socks.interpretResponseBuffer(buffer);
+      if (response.reply != Socks.Reply.SUCCEEDED) {
+        return Promise.reject(response);
       }
       return Promise.resolve(connection);
     });
--- a/src/integration/socks-echo/socks-echo.integration.spec.ts
+++ b/src/integration/socks-echo/socks-echo.integration.spec.ts
@@ -6,8 +6,6 @@
 // Integration test for the whole proxying system.
 // The real work is done in the Freedom module which performs each test.
 describe('proxy integration tests', function() {
-  var testModule :any;
-
   var testStrings = [
     'foo',
     'bar',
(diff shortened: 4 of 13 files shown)


Gruntfile.coffee
```diff
@@ -336,7 +336,7 @@ module.exports = (grunt) ->
             {path: 'build/tcp/tcp.js', include: false}
             {path: 'build/integration/*/integration.*', include: false}
           ]
-          keepBrowser: false
+          keepBrowser: true
```
willscott: probably don't want this except locally?

bemasc: Whoops.  Fixed.  Thanks!

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185)

---

I have a bunch of comments and suggestions (sorry!) but otherwise looking good!

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 119 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHXMDPms79NJNZ6Dv0)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L119)):</sup>
It's a little messy that your `beforeEach` will run before this, too -- this suggests that it's not really something that should be run before every test anymore. I would suggest renaming `beforeEach` and adding a boolean argument which each function can call.

What do you think?

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 138 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHZI0ahaKClC2ccYhA)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L138)):</sup>
So, what exactly are you trying to test here and in the next test: that we can make a connection via the SOCKS proxy to an address other than localhost?

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 170 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHZWasSw2Nh-PgsTek)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L170)):</sup>
So, I'm not surprised to come across a DNS server which filters non-routable addresses.

We could have one test which fails iff a connection can be made to an address which resolves to a non-routable IP.

That this will yield false negatives on some/corporate networks suggests that this is a better candidate for a unit test. There, we can mock in all sorts of crazy DNS results.

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 205 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgH_dyEctmJksyLTeW4)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L205)):</sup>
This is a good thing to test but how do you know there is no service running on this port? :-)

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 214 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgH_j5FpeIcGHufl-go)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L214)):</sup>
Nice.

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 223 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgH_khy1ZVPp0CUCJXT)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L223)):</sup>
Nice. I wonder if OSes have varying behaviour here...but definitely something we should be testing.

---

<sup>**[src/socks-common/socks-headers.d.ts, line 42 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHRZFZcHcxDIvVCKfb)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/socks-common/socks-headers.d.ts#L42)):</sup>
So, great to see a proper `composeResponse` function to go with `composeRequest`.

I think we can do better with regards to naming, though! Right now, it's not clear what's the difference between `Reply` and `Response`. Also, `Reply` has a field of type `Response` named `replyField`. :-)
1. Reading the RFC, our current `Response` type should be renamed `Reply` (https://tools.ietf.org/html/rfc1928#section-6)
2. Your new type can be named `Response`
3. The functions can be named `composeResponseBuffer` and `interpretResponseBuffer`.

Sorry I know you've used `Reply` in a bunch of places but I think `Response` is better, especially given that the RFC has a special use for the term Reply.

---

<sup>**[src/tcp/tcp.ts, line 245 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHG9suzUyOw8ASelvO)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/tcp/tcp.ts#L245)):</sup>
It seems totally reasonable to add a way to query the client socket's address (oops, never spotted that before). It's a pity we have to change the signature on this function, though -- FWIW, these "convenience" combined do/get functions have always bothered me for precisely this reason.

If we had had:

``` typescript
public onceConnected :Promise<void>;
public getRemoteEndpoint :Promise<Net.Endpoint>;
```

Then you could have just added the following function, without any disruption to current callers and without the need to define a new type:

``` typescript
public getLocalEndpoint :Promise<Net.Endpoint>;
```

Without necessarily doing it in this pull request, what are your thoughts on that approach?

---

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185)

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 126 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHXMDPms79NJNZ6Dv0)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L126)):</sup>
Done.

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 138 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHZI0ahaKClC2ccYhA)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L138)):</sup>
Yes.  I've changed this to do a trivial HTTP fetch to a nonexistent URL on uproxy.org, in order to get a 404 response that we can check.

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 171 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHZWasSw2Nh-PgsTek)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L171)):</sup>
Done.  Unfortunately, it still doesn't work yet; currently the SOCKS server just never responds if DNS lookup fails.   I'm working on that for a separate pull request.

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 210 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgH_dyEctmJksyLTeW4)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L210)):</sup>
1023 is a special reserved port.  Nothing is allowed to listen on it.  I've added a comment noting this.

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 219 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgH_j5FpeIcGHufl-go)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L219)):</sup>
Thanks.  I've opened https://github.com/uProxy/uproxy/issues/800 to get this fixed.

---

<sup>**[src/socks-common/socks-headers.d.ts, line 42 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHRZFZcHcxDIvVCKfb)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/socks-common/socks-headers.d.ts#L42)):</sup>
Done.

---

<sup>**[src/tcp/tcp.ts, line 245 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHG9suzUyOw8ASelvO)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/tcp/tcp.ts#L245)):</sup>
I think it depends in part on how much information can be extracted from a failed socket.  Consider the case where the socket is instructed to connect to a DNS name that successfully resolves to an IP address, but then the server refuses the connection.  In the current interface, we can never learn the IP address.  If the underlying socket actually allows this, then having separate getters would be better, to get info from failed sockets.

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185)

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 17 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgMXVmFgWtHuKNP6s7q)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/85f083403d57204f222dfd6f5ecaebccb0543f1d/src/integration/socks-echo/socks-echo.integration.spec.ts#L17)):</sup>
should the arg be optional?

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 141 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHZI0ahaKClC2ccYhA)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L141)):</sup>
OK. Then I would suggest renaming the test to something like "fetch from non-localhost address" or similar (accessing uproxy.org is the means rather than the end).

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 199 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgH_dyEctmJksyLTeW4)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L199)):</sup>
Ah! I didn't know. Cool!

---

<sup>**[src/socks-common/socks-headers.d.ts, line 42 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHRZFZcHcxDIvVCKfb)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/socks-common/socks-headers.d.ts#L42)):</sup>
Much better. Thanks!

---

<sup>**[src/socks-common/socks-headers.ts, line 86 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgMcQZAxoQE9tynLqv8)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/85f083403d57204f222dfd6f5ecaebccb0543f1d/src/socks-common/socks-headers.ts#L86)):</sup>
So, the TypeScript-foo is kind of neat but my instinct is that the asserts performed by these functions could just be added to `interpretRexxxBuffer` and `composeRexxxBuffer` -- they could raise an exception when asked to encode or decode an invalid `ArrayBuffer` or `Object`. The `interpretRexxxBuffer` methods already perform some sanity checking.

You might achieve this with the introduction of an `isValidResponse` function, or similar (a functional programming-like approach, I guess).

What do you think?

---

<sup>**[src/tcp/tcp.ts, line 245 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHG9suzUyOw8ASelvO)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/tcp/tcp.ts#L245)):</sup>
I don't completely follow but I think you're giving another reason for why separating the `do`-ing from the `get`-ting would help :-)

I don't want to derail this pull request in the pursuit of API perfection but I am strongly in favour of a slight redesign here soon.

---

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185)

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 17 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgMXVmFgWtHuKNP6s7q)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/85f083403d57204f222dfd6f5ecaebccb0543f1d/src/integration/socks-echo/socks-echo.integration.spec.ts#L17)):</sup>
It's currently optional.  Would you prefer it to be non-optional, and have each test explicitly pass `true` or `false`?

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 141 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHZI0ahaKClC2ccYhA)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/integration/socks-echo/socks-echo.integration.spec.ts#L141)):</sup>
Done.

---

<sup>**[src/socks-common/socks-headers.ts, line 86 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgMcQZAxoQE9tynLqv8)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/85f083403d57204f222dfd6f5ecaebccb0543f1d/src/socks-common/socks-headers.ts#L86)):</sup>
This is not related to `compose*` or `interpret*`.   These parsers are used by RtcToNet to read untrusted JSON objects sent by SocksToRtc, not to read bytes in the Socks protocol.

I think a parser function is preferable because it avoids having the usable-but-not-sanitized object available in SocksToRtc and RtcToNet (and because it removes the need for the caller to typecast if `isValid*` returns true).

---

<sup>**[src/tcp/tcp.ts, line 245 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHG9suzUyOw8ASelvO)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/tcp/tcp.ts#L245)):</sup>
Yes.  I just checked (the Chrome API docs)[https://developer.chrome.com/apps/socket#method-getInfo], which I think say that `getInfo` will still provide useful info even if the socket doesn't actually connect successfully. 

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185)

---

<sup>**[src/integration/socks-echo/socks-echo.integration.spec.ts, line 17 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgMXVmFgWtHuKNP6s7q)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/85f083403d57204f222dfd6f5ecaebccb0543f1d/src/integration/socks-echo/socks-echo.integration.spec.ts#L17)):</sup>
Fine as-is, I guess.

---

<sup>**[src/socks-common/socks-headers.ts, line 86 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgMcQZAxoQE9tynLqv8)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/85f083403d57204f222dfd6f5ecaebccb0543f1d/src/socks-common/socks-headers.ts#L86)):</sup>
I get that but by factoring out some `isValid` functions you could:
- sanitise the untrusted JSON (what you care about)
- simplify and improve the `interpre*` and `parseRequestBuffer` and `parseResponseBuffer` functions (bonus)
- eliminate the need to introduce `parseRequest` and `parseResponse` functions (more bonus)

Basically, I think `parseRequest` and `parseResponse` are unnecessarily specific to crafting and parsing JSON.

---

<sup>**[src/tcp/tcp.ts, line 245 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgHG9suzUyOw8ASelvO)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/76cf33ca9bf5f434d752fb2cc84e1b25e1f61189/src/tcp/tcp.ts#L245)):</sup>
I filed this:
https://github.com/uProxy/uproxy/issues/803

---

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185)

---

<sup>**[src/socks-common/socks-headers.ts, line 86 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgMcQZAxoQE9tynLqv8)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/85f083403d57204f222dfd6f5ecaebccb0543f1d/src/socks-common/socks-headers.ts#L86)):</sup>
Done.

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185)

---

Thanks. I think this is a lot easier for the next person who comes along.

:+1: once you're happy the checking code is all factored out nicely.

---

<sup>**[src/socks-common/socks-headers.ts, line 107 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgN3NL6vawwJ54KvEx8)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/09861673b76f8d33d1295bbd891d4d30b46cf4a0/src/socks-common/socks-headers.ts#L107)):</sup>
So, now that you have this you can greatly simplify `interpretResponseBuffer` and `interpretRequest` -- replace all the checks with a single call to this function (or its `Request` counterpart) before returning to the user. Make sense? (you might want this to raise errors, to help with debugging).

Does that make sense?

---

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185)

---

<sup>**[src/socks-common/socks-headers.ts, line 107 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/185#-JgN3NL6vawwJ54KvEx8)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/09861673b76f8d33d1295bbd891d4d30b46cf4a0/src/socks-common/socks-headers.ts#L107)):</sup>
I only see one check that could be replaced with such a call in `interpretRequest` and `interpretResponse`, and I don't see how the replacement would help.  Could you be more specific?

---

PR #188, opened 2015-01-29
move to uproxy-lib 18, with loggingprovider

trevj:
OK, so, yeah. This updates the repo to use uproxy-lib 18.

This has the new loggingprovider stuff so it's mostly a matter of updating the sample apps to add loggingprovider as a dependency and updating calls to `setConsoleFilter`.

@willscott: Does this look about right to you? Do you spot any unnecessary stuff in here? It's quite a big diff, would appreciate any suggestions you might have on trimming it.

@iislucas: I know you are working on some improvements to mocking with the move to require but hopefully this is not unacceptable.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/188)

### DIFF
--- a/src/churn/mocks.js
+++ b/src/churn/mocks.js
@@ -2,9 +2,13 @@
 // We do this in a non-TypeScript file because the ambient module declaration
 // prevents us creating any variable called freedom in TypeScript-land.
 var freedom = {
-  core: jasmine.createSpy().and.returnValue(
-    jasmine.createSpyObj('core', ['getLogger'])),
   churn: jasmine.createSpy().and.returnValue(
     jasmine.createSpyObj('churn', ['providePromises']))
 };
-
+freedom['core'] = function() {
+  return {
+    'getLogger': function() {
+      return jasmine.createSpyObj('logger', ['then']);
+    }
+  };
+};
--- a/src/churn/mocks.js
+++ b/src/churn/mocks.js
@@ -1,9 +1,10 @@
 // Create a mock instance of Freedom.
 // We do this in a non-TypeScript file because the ambient module declaration
 // prevents us creating any variable called freedom in TypeScript-land.
-var freedom = jasmine.createSpyObj('freedom', ['core.log']);
-freedom.churn = jasmine.createSpy().and.returnValue(
-    jasmine.createSpyObj('churn', ['providePromises']));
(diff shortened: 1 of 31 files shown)


src/churn/mocks.js
```diff
@@ -1,9 +1,14 @@
 // Create a mock instance of Freedom.
 // We do this in a non-TypeScript file because the ambient module declaration
 // prevents us creating any variable called freedom in TypeScript-land.
-var freedom = jasmine.createSpyObj('freedom', ['core.log']);
-freedom.churn = jasmine.createSpy().and.returnValue(
-    jasmine.createSpyObj('churn', ['providePromises']));
-freedom['core.console'] = function() {
-  return window.console;
+var freedom = {
```
willscott, line 4: we also have freedom mocks in uproxy-lib. is there a way to have them centralized in one place?

src/echo/freedom-module.ts
```diff
@@ -2,6 +2,8 @@
 /// <reference path='../logging/logging.d.ts' />
 /// <reference path='../freedom/typings/freedom.d.ts' />
 
+freedom['loggingprovider']().setConsoleFilter(['*:D']);
```
willscott, line 5: this is default behavior per https://github.com/uProxy/uproxy-lib/blob/dev/src/loggingprovider/loggingprovider.ts#L33  Why needed?

willscott:
Looks generally good to me - only real thought on trimming would be trying to centralize the mock to one shared location, and maybe removing the 'loggingprovider' interface from the demos that don't set a meaningful console filter.

src/churn/mocks.js
trevj, line 4: There is, and @iislucas is working on this as part of his big require change. So, we can benefit from that when the time comes to update this repo.

src/echo/freedom-module.ts
trevj, line 5: It's often useful to adjust the logging level when working on new features or debugging an issue, so this is an example of how to do it.

trevj:
Thanks for looking!

@iislucas I'll merge this if you don't have any objections after taking a quick look over?

bemasc:
:+1:

I don't really know what's going on here, but if it works, it's fine by me.

PR #189, opened 2015-01-29
bin/make-data.sh now makes text files.

mollyling:
Instead of creating binary files that seem to decode incorrectly in the benchmark (due to mime-type issues, AFAICT), make text files instead.

The script starts with the same binary files, then hex-dumps that, and then pulls the appropriate amount of data out of the hex dump.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/189)

### DIFF
--- a/bin/make-data.sh
+++ b/bin/make-data.sh
@@ -3,7 +3,10 @@ mkdir data
 pushd data
 for i in 1024 2048 4096 8192 1048576 2097152 16777216
 do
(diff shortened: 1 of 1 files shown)


mollyling:
This should close https://github.com/uProxy/uproxy/issues/770 once merged.

trevj:
:+1:

PR #193, opened 2015-02-05
Add integration testing for churn by running the existing integration tests twice

bemasc:
This works ... but it causes both the new tests _and_ the existing tests to crash the renderer.  (Filed as https://crbug.com/454578)

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/193)


bemasc:
@trevj and @willscott, PTAL, and also consider the crasher issue.

Also, if anyone knows how to make github or Reviewable follow diffs across file-copies, let me know.

src/integration/socks-echo/integration.ts
```diff
@@ -5,6 +5,8 @@
 /// <reference path="../../tcp/tcp.d.ts" />
 /// <reference path="../../webrtc/peerconnection.d.ts" />
 
+freedom['loggingprovider']().setConsoleFilter(['*:D']);
```
willscott, line 8: maybe try a higher logging level?

Gruntfile.coffee
```diff
@@ -18,6 +18,31 @@ FILES =
...
+  'build/simple-transformers/*.js'
+  'build/regex2dfa/*.js'
+  'build/utransformers/**/*.js'
+  'build/churn-pipe/churn-pipe.js'
+  'build/churn-pipe/freedom.json'
+  'build/churn/churn.js'
+  'build/webrtc/*.js'
+  # spec_base.js is listed first so that it can be used by spec.js
+  'build/integration/**/*.integration.spec_base.js'
+  'build/integration/**/*.js'
```
willscott, line 42: doesn't this subsume the above rule as well?

bemasc, line 42: Yes.  Hence the comment.  spec_base.js must be included first in main.html, and this list sets the ordering, so they must be listed separately.

Evidently the comment was not clear enough.  Improved wording suggestions welcome.

src/integration/socks-echo/integration.ts
bemasc, line 8: You don't think debug logging is appropriate for tests?

trevj, line 8: Definitely keep debug for these tests. Thanks for adding, this is useful for others to quickly change the logging level, too.

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/193)

---

I can't access the Chromium bug so can't comment on that right now but otherwise this looks fine and it's GREAT to have churn in our integration tests. Many thanks!

:+1:

---

PR #195, opened 2015-02-12
Fix test flakiness and crashes

bemasc:
Flakiness was due to the tests not handling Nagling.  Crashes were due to not cleaning up copies of the freedom app, leading to huge numbers of web workers hanging around.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/195)


trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/195)

---

Wow. :+1:

---

PR #200, opened 2015-02-18
package.json sweep

trevj:
Mostly to get the latest freedom-for-chrome but just generally it's been a while since we did this.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/200)


iislucas:
:+1:

PR #201, opened 2015-02-19
instrumentation for RtcToNet

trevj:
This adds the notion of "snapshots" to `RtcToNet`. These are for instrumentation, so that we can visualise the behaviour of the proxy over time.

Snapshots are dumped every 5 seconds, as JSON in the format:

``` json
{  
  "sessions":[  
    {  
      "name":"c0",
      "channel":{  
        "sent":19289,
        "received":61599,
        "buffered":0,
        "queue":{  
          "size":0,
          "handling":true
        }
      },
      "socket":{  
        "sent":61599,
        "received":19289,
        "queue":{  
          "size":0,
          "handling":true
        }
      }
    },
    {  
      "name":"c1",
      "channel":{  
        "sent":29504,
        "received":5842,
        "buffered":0,
        "queue":{  
          "size":0,
          "handling":true
        }
      },
      "socket":{  
        "sent":5842,
        "received":29504,
        "queue":{  
          "size":0,
          "handling":true
        }
      }
    }
  ]
}
```

@mollyling: Is this something you can work with to make graphs? What would you change?

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/201)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-networking",
   "description": "uProxy's networking library: SOCKS5 over WebRTC",
-  "version": "5.2.6",
+  "version": "5.3.0",
   "repository": {
     "type": "git",
     "url": "https://github.com/uProxy/uproxy-networking"
--- a/src/rtc-to-net/rtc-to-net.d.ts
+++ b/src/rtc-to-net/rtc-to-net.d.ts
@@ -69,6 +69,7 @@ declare module RtcToNet {
     public onceClosed :Promise<void>;
     public close :() => void;
     public handleSignalFromPeer :(signal:WebRtc.SignallingMessage) => void;
+    public initiateSnapshotting :() => void;
     public toString :() => string;
   }
   class Session {
--- a/src/rtc-to-net/rtc-to-net.d.ts
+++ b/src/rtc-to-net/rtc-to-net.d.ts
@@ -25,8 +25,6 @@ declare module RtcToNet {
     bytesSent :number;
     // Total number of bytes received since the connection was created.
     bytesReceived :number;
-    // True iff the socket is paused right now.
-    paused :boolean;
     // Data received from the peer.
     dataIn :HandlerQueueSnapshot;
   }
--- a/src/rtc-to-net/rtc-to-net.d.ts
+++ b/src/rtc-to-net/rtc-to-net.d.ts
@@ -15,29 +15,29 @@ declare module RtcToNet {
   // for later analysis.
   interface HandlerQueueSnapshot {
     // Number of objects waiting to be handled right now.
-    length :number;
+    size :number;
     // True iff there is a handler attached right now.
     handling :boolean;
   }
 
   interface SocketSnapshot {
     // Total number of bytes sent since the connection was created.
-    bytesSent :number;
+    sent :number;
     // Total number of bytes received since the connection was created.
-    bytesReceived :number;
+    received :number;
     // Data received from the peer.
-    dataIn :HandlerQueueSnapshot;
+    queue :HandlerQueueSnapshot;
   }
 
   interface DataChannelSnapshot {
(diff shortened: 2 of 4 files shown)


trevj:
Part of:
https://github.com/uProxy/uproxy/issues/946

trevj:
BTW, this is kinda janky code -- I don't want to put too much work into a better solution until we're sure this kinda approach is useful. This gets the numbers into the log, and little else.

mollyling:
Can we keep in sockets that were closed on error for some time?  To see what their stats were?  Also, the time between the last-sent byte and the last-received byte would be useful as a rough latency metric.

trevj:
1. Not easily because the session is removed from the map rightaway and I don't really want to mess with that code. However, there is an entry in the log for when a datachannel or socket closes with an error.
2. A simple counter like this should allow us graph rate over time, right? Do you think "time since last byte sent or received" adds anything?

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/201)

Reviewed files:
- src/rtc-to-net/rtc-to-net.d.ts @ r1

---

Looks helpful to me! 
One comment/idea. Otherwise :+1:

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 169 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/201#-JiZN1sevQ2QSWvLRcNG)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/e7cb0edb5e5ed8152d52ad634f0953e97d5b4029/src/rtc-to-net/rtc-to-net.ts#L169)):</sup>
Might be better to make this public so it can be called from the console, and then the code doesn't have to be edited to enable it? 

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/201)

---

Thanks!

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 169 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/201#-JiZN1sevQ2QSWvLRcNG)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/e7cb0edb5e5ed8152d52ad634f0953e97d5b4029/src/rtc-to-net/rtc-to-net.ts#L169)):</sup>
Ooh yes, good idea. Done.

---

mollyling:
Ok, if we can extract the killed sockets later (and mark them up in our data) that's just as good.

I was hoping to get some latency numbers out of this.  So if we have both the time of last byte sent, and time of last byte received, we've got the rough request->response latency by subtraction.

trevj:
Update: Seeing weird issues with this when the system is under heavy load...from speaking with @bemasc it seems we can be a lot smarter about how we use handler queues. Gonna hold off on this until we have a better understanding since analysing this when the system is under load was kinda the whole point :-)

trevj:
Update: output the data channel's actual `bufferedAmount`.

trevj:
OK, I merged in recent changes to dev. The system now outputs snapshots even when it's under load.

Otherwise unchanged...PTAL!

mollyling:
Generally, I'm quite happy with this. In a future cl, three small things:
- some time info to calculate latencies and identify their components
- a way to alter the snapshot interval. Sometimes a higher rate is desirable, even if it adds overhead.
- and can we automate control of enable/disable the snapshotting? So we can have a load test at different levels of load, sampling rates, and collect separate data for each.

LGTM
:+1:

trevj:
Thanks!

@mollyling a couple of things:
- Could you file an issue describing the latencies you want?
- At first I guess we'll just copy and paste the JavaScript console to a file and process it that way so there'll be manual steps involved regardless (enabling, snapshot interval, etc.) Let's consider automation when we have a better idea how well all this will work.

PR #202, opened 2015-02-20
improve TCP connection disconnect logging

trevj:
Related issue:
https://github.com/uProxy/uproxy/issues/964

I don't think `RtcToNet` really cares why the connection was closed, so let's stop logging it. Since I don't see any issues with proxy behaviour after encountering lots of these messages, let's just log them all as INFO.

I think it _would_ be useful to improve the logging in `tcp.ts`. Having done that, I see several of these warnings:

```
Got onDisconnect in closed state
```

I will update the issue and continue to investigate. Might be harmless, might not.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/202)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-networking",
   "description": "uProxy's networking library: SOCKS5 over WebRTC",
-  "version": "5.2.2",
+  "version": "5.2.3",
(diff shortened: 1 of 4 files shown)


bemasc:
:+1:

PR #203, opened 2015-02-23
improve responsiveness of SocksToRtc and RtcToNet queue processing

trevj:
So, inspired by this pull request:
https://github.com/uProxy/uproxy-lib/pull/126/files

This greatly improves the responsiveness of the proxy under high load. Specifically, with this change the proxy is sitll responsive while there is a fast download taking place, e.g.:
1. start curl-ing a large file
2. try to browse "normal" web pages with a browser

This does not currently work. General idea is to yield after each fetch from the socket and data channel's queue. More info here:
https://github.com/uProxy/uproxy/issues/967

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/203)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-networking",
   "description": "uProxy's networking library: SOCKS5 over WebRTC",
-  "version": "5.2.3",
+  "version": "5.2.4",
   "repository": {
     "type": "git",
     "url": "https://github.com/uProxy/uproxy-networking"
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -404,7 +404,7 @@ module RtcToNet {
 
     // Sends a packet over the data channel.
     // Invoked when a packet is received over the TCP socket.
-    private readFromSocket_ = (data:ArrayBuffer) : void => {
+    private sendOnChannel_ = (data:ArrayBuffer) : void => {
       log.debug('%1: socket received %2 bytes', [
           this.longId(),
           data.byteLength]);
@@ -417,7 +417,7 @@ module RtcToNet {
 
     // Sends a packet over the TCP socket.
     // Invoked when a packet is received over the data channel.
-    private readFromChannel_ = (data:WebRtc.Data) : void => {
+    private sendOnSocket_ = (data:WebRtc.Data) : void => {
       if (!data.buffer) {
         log.error('%1: received non-buffer data from datachannel', [
             this.longId()]);
@@ -450,7 +450,7 @@ module RtcToNet {
     // been successfully established.
     private startReadLoops_ = () : void => {
       var socketReadLoop = (data:ArrayBuffer) => {
-        this.readFromSocket_(data);
+        this.sendOnChannel_(data);
         Session.nextTick_(() => {
           this.tcpConnection_.dataFromSocketQueue.setSyncNextHandler(
               socketReadLoop);
@@ -460,7 +460,7 @@ module RtcToNet {
           socketReadLoop);
 
       var channelReadLoop = (data:WebRtc.Data) : void => {
-        this.readFromChannel_(data);
+        this.sendOnSocket_(data);
         Session.nextTick_(() => {
           this.dataChannel_.dataFromPeerQueue.setSyncNextHandler(
               channelReadLoop);
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -25,6 +25,8 @@ module RtcToNet {
 
   // The |RtcToNet| class holds a peer-connection and all its associated
   // proxied connections.
+  // TODO: Extract common code for this and SocksToRtc:
+  //         https://github.com/uProxy/uproxy/issues/977
   export class RtcToNet {
     // Configuration for the proxy endpoint. Note: all sessions share the same
     // (externally provided) proxyconfig.
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -268,7 +268,7 @@ module RtcToNet {
           }
           throw e;
         });
-      this.onceReady.then(this.linkTcpAndPeerConnectionData_);
+      this.onceReady.then(this.startReadLoops_);
 
       this.onceReady.catch(this.fulfillStopping_);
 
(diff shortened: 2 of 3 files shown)


trevj:
@iislucas you might also be interested.

bemasc:
Do we have any understanding of why this should help?  Were you testing with a version of uproxy-lib that includes Friday's queueing fix?

trevj:
My understanding is that `Queue.processQueue_` occupies the thread until the queue is exhausted, which is an extremely bad idea if your queue is large.

I was testing with that change but without this change the proxy is still extremely unresponsive while under heavy load. I'm sure the fix to uproxy-lib is good but this also seems to be necessary.

bemasc:
This is true ... but all it's doing is moving JS objects from one JS queue into another.  It should take approximately zero time (it's just a pointer operation).  It's not like we're blocking the thread to do I/O, or even a memcpy.

In contrast, moving back to the mainloop after every packet represents an increase in CPU load on that thread, possibly nontrivial.  As such, my expectation is that this change should make performance worse, not better.

I'd be a lot happier with this change if I had a clearer understanding of why it helps.

bemasc:
@trevj and I spent a while discussing this and we have a tentative understanding.  Essentially, the pull request to make datachannel.ts slow down at the buffer limit did not actually work.  If it had worked, this would not be necessary: RtcToNet would pile all incoming data directly into the DataChannel queue, which would only pull it out one message at a time while checking getBufferedAmount, which should effectively give priority to the connections that send the least data.  Instead, we see the opposite behavior.

We don't know why this didn't work, but in the interest of unblocking closed beta, we can land this patch, and I will try to investigate why that code doesn't behave as expected.

This patch introduces a sort of bucket brigade, with two successive points (in one thread) where one queue is drained into another, one message at a time, asynchronously (using setTimeout).  This should not be necessary, and perhaps it can be simplified (e.g. https://github.com/uProxy/uproxy/issues/979) once we understand why datachannel.ts is not behaving as expected.

:+1:

iislucas:
What do you think of abstracting `setTimeout(..., 0)` into a named function? I think it'll read better and be easier to adapt to different implementations.

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203)

Reviewed files:
- package.json @ r1

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 407 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-JisjrOJVTWfXvEn0yx6)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5b8c9795eb4741e17ff927cc24a0f10f20f57ef9/src/rtc-to-net/rtc-to-net.ts#L407)):</sup>
I think this should be called `sendOnDataChannel_` it doesn't read anything from the socket.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 420 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-Jisjx0TQ4CSPmT7oZjY)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5b8c9795eb4741e17ff927cc24a0f10f20f57ef9/src/rtc-to-net/rtc-to-net.ts#L420)):</sup>
I think this should be called `sendOnTcpSocket_`. Same goes for names in socks-to-rtc

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 451 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-Jisk0pKhX6BdTE3R4Fi)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5b8c9795eb4741e17ff927cc24a0f10f20f57ef9/src/rtc-to-net/rtc-to-net.ts#L451)):</sup>
I think I prefer the old name; this doesn't do anything to start reading data; it just connects the two data sending pipes. 

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 314 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-JiskSATJmlIhVKKCfcv)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5b8c9795eb4741e17ff927cc24a0f10f20f57ef9/src/socks-to-rtc/socks-to-rtc.ts#L314)):</sup>
Again, I think I prefer the old name (Reading happens independently of this call)

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203)

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203)

---

I also factored out a `nextTick_` method, as you suggested. Growing weight for tackling:
https://github.com/uProxy/uproxy/issues/977

Just one question for you. PTAL.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 407 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-JisjrOJVTWfXvEn0yx6)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5b8c9795eb4741e17ff927cc24a0f10f20f57ef9/src/rtc-to-net/rtc-to-net.ts#L407)):</sup>
Good point! Renamed to `sendOnSocket_` and `sendOnChannel_` (in the interest of brevity).

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 420 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-Jisjx0TQ4CSPmT7oZjY)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5b8c9795eb4741e17ff927cc24a0f10f20f57ef9/src/rtc-to-net/rtc-to-net.ts#L420)):</sup>
See above.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 451 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-Jisk0pKhX6BdTE3R4Fi)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5b8c9795eb4741e17ff927cc24a0f10f20f57ef9/src/rtc-to-net/rtc-to-net.ts#L451)):</sup>
So, it _does_ initiate reading via the calls to `setSyncNetxHandler` and arguably it's _more_ of a loop than before!

I would be okay with `linkSocketAndChannel_`; what do you think?

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 314 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-JiskSATJmlIhVKKCfcv)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5b8c9795eb4741e17ff927cc24a0f10f20f57ef9/src/socks-to-rtc/socks-to-rtc.ts#L314)):</sup>
See above.

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203)

Reviewed files:
- src/rtc-to-net/rtc-to-net.ts @ r2
- src/socks-to-rtc/socks-to-rtc.ts @ r2

---

OK, I like `linkSocketAndChannel_` better. 

I'm actually confused as to why the timeouts are needed; it looks to me like they don't do anything... lets have a chat later and I'll try to understand a bit more what's happening here. 

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 451 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-Jisk0pKhX6BdTE3R4Fi)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5b8c9795eb4741e17ff927cc24a0f10f20f57ef9/src/rtc-to-net/rtc-to-net.ts#L451)):</sup>
`linkSocketAndChannel_` sounds good to me. 

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 512 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-JiwIDySL0XmAljxHM0v)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/4c1d7f02c50313452f0b807b2e4b52d5465d296c/src/socks-to-rtc/socks-to-rtc.ts#L512)):</sup>
I'd add a TODO to say that this and the version above should be moved into an appropriate util library, probably with the `keepAlive` function to stop apps suspending; once we have one more related thing to put in that to be named module about runtime management. :) 

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203)

---

OK, as discussed I was able to replace `nextTick_` with promise chaining the calls to `setSyncNextHandler`. Great find, thanks.

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 451 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-Jisk0pKhX6BdTE3R4Fi)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5b8c9795eb4741e17ff927cc24a0f10f20f57ef9/src/rtc-to-net/rtc-to-net.ts#L451)):</sup>
Done.

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 512 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203#-JiwIDySL0XmAljxHM0v)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/4c1d7f02c50313452f0b807b2e4b52d5465d296c/src/socks-to-rtc/socks-to-rtc.ts#L512)):</sup>
I added a TODO to SocksToRtc and RtcToNet linking to this issue:
https://github.com/uProxy/uproxy/issues/977

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203)

---

Hmm, not so fast. With promise chaining, responsiveness is improved over HEAD. However, it's not as good as with `setTimeout` and performance is significantly worse.

I see some variation between runs but it's clear to me that the promise chaining approach uses much more CPU and often de-generates into something approaching the behaviour of HEAD.

How I tested:
1. inspired by my instrumentation pull request, I hack `RtcToNet` to output `snapshot!` to the console, on a 5000ms `setTimeout`
2. patch that into this pull request (call this a)
3. patch that into the previous `setTimeout` incarnation of this pull request (call this b)

Now, for both a and b, I:
1. leave `curl` downloading from a fast server at a `--rate-limit`-ed 50k/sec
2. record when `snapshot!` appears on the console
3. try to download some other, small, pages concurrent to the curl
4. record my results

For a (promise chaining):
- curl starts at 50k/sec but drops to ~20k sec before recovering to 50k/sec...
- ~200-300% CPU use
- small download is responsive
- snapshots are very irregular:

```
RtcToNet I [Feb 24 14:13:34.991] snapshot
RtcToNet I [Feb 24 14:13:39.993] snapshot
RtcToNet I [Feb 24 14:13:44.992] snapshot
RtcToNet I [Feb 24 14:13:52.517] snapshot
RtcToNet I [Feb 24 14:14:00.622] snapshot
RtcToNet I [Feb 24 14:14:14.489] snapshot
RtcToNet I [Feb 24 14:14:17.501] snapshot
RtcToNet I [Feb 24 14:14:20.723] snapshot
```

For b (`setTimeout`):
- curl is always downloading at 50k/sec
- ~150-200% CPU use
- 

```
RtcToNet I [Feb 24 14:11:49.518] snapshot
RtcToNet I [Feb 24 14:11:54.520] snapshot
RtcToNet I [Feb 24 14:11:59.702] snapshot
RtcToNet I [Feb 24 14:12:05.164] snapshot
RtcToNet I [Feb 24 14:12:09.551] snapshot
RtcToNet I [Feb 24 14:12:15.175] snapshot
RtcToNet I [Feb 24 14:12:22.992] snapshot
RtcToNet I [Feb 24 14:12:26.470] snapshot
RtcToNet I [Feb 24 14:12:29.580] snapshot
RtcToNet I [Feb 24 14:12:34.599] snapshot
RtcToNet I [Feb 24 14:12:39.588] snapshot
RtcToNet I [Feb 24 14:12:44.589] snapshot
RtcToNet I [Feb 24 14:12:49.589] snapshot
RtcToNet I [Feb 24 14:12:54.590] snapshot
RtcToNet I [Feb 24 14:12:59.590] snapshot
RtcToNet I [Feb 24 14:13:04.591] snapshot
RtcToNet I [Feb 24 14:13:09.593] snapshot
```

This was on a "good" "promise chaining" run. On several runs, I saw download speed fall below 10k/sec with ~300% CPU use. Even without automatic, reproducible, benchmarking it's clear to me that we should stick with `setTimeout`. So, I've reverted to the previous commit and preserved the promise chaining approach here for now:
https://github.com/uProxy/uproxy-networking/commit/50a4ca3f2122d2b75621d87b07a71ea4e261a1c7

PTAL!

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/203)

Reviewed files:
- src/rtc-to-net/rtc-to-net.ts @ r3
- src/socks-to-rtc/socks-to-rtc.ts @ r3

---

OK, I wrote a little code to demo behavior of promises w.r.t. timeouts: 

```
var count_ = 0;
function makeFulfilledPromised() {
    console.log(++count_ + ': makeFulfilledPromised.');
    return new Promise(function (F, R) {
        console.log(++count_ + ': makeFulfilledPromised:F');
        F();
    });
}
function fooPromiseLoop() {
    console.log(++count_ + ': fooPromiseLoop start.');
    setTimeout(function () {
        console.log(++count_ + ': init setTimeout.');
    }, 0);
    var promiseState = makeFulfilledPromised();
    console.log(++count_ + ': loopstart.');
    for (var i = 0; i < 4; i++) {
        console.log(++count_ + ': fooPromiseLoop start: ' + i);
        setTimeout(function () {
            console.log(++count_ + ': setTimeout: ' + i);
        }, 0);
        promiseState = promiseState.then(function () {
            console.log(++count_ + ': promiseState.then: ' + i);
            return makeFulfilledPromised().then(function () {
                console.log(++count_ + ': makeFulfilledPromised.then: ' + i);
            });
        });
        console.log(++count_ + ': fooPromiseLoop end: ' + i);
    }
    console.log(++count_ + ': loopend.');
}
```

which gives output: 

```
> fooPromiseLoop();
fooPromiseLoop();
VM87:11 1: fooPromiseLoop start.
VM87:4 2: makeFulfilledPromised.
VM87:6 3: makeFulfilledPromised:F
VM87:16 4: loopstart.
VM87:18 5: fooPromiseLoop start: 0
VM87:31 6: fooPromiseLoop end: 0
VM87:18 7: fooPromiseLoop start: 1
VM87:31 8: fooPromiseLoop end: 1
VM87:18 9: fooPromiseLoop start: 2
VM87:31 10: fooPromiseLoop end: 2
VM87:33 11: loopend.
VM87:23 12: promiseState.then.1: 3
VM87:26 13: promiseState.then.2: 3
VM87:4 14: makeFulfilledPromised.
VM87:6 15: makeFulfilledPromised:F
VM87:28 16: makeFulfilledPromised.then: 3
VM87:23 17: promiseState.then.1: 3
VM87:26 18: promiseState.then.2: 3
VM87:4 19: makeFulfilledPromised.
VM87:6 20: makeFulfilledPromised:F
VM87:28 21: makeFulfilledPromised.then: 3
VM87:23 22: promiseState.then.1: 3
VM87:26 23: promiseState.then.2: 3
VM87:4 24: makeFulfilledPromised.
VM87:6 25: makeFulfilledPromised:F
VM87:28 26: makeFulfilledPromised.then: 3
undefined
VM87:13 27: init setTimeout.
VM87:20 28: setTimeout: 3
VM87:20 29: setTimeout: 3
VM87:20 30: setTimeout: 3
```

Notice that execution of `then` functions happens on the same execution cycle as the resolved promise, just after all the other code. Generally, seems that promise's don't give clearly defined timings for when stuff gets executed, so when using async behaviour, it's important to imagine all possible execution models. 

So all of that said, I see now why nextTick_ is needed. 

Big :+1: for figuring out this solution. Nice work

---

PR #204, opened 2015-02-24
eliminate spurious 'Got onDisconnect in closed state' warnings

trevj:
So, main thing is to remove these warnings from `tcp.ts`:
https://github.com/uProxy/uproxy/issues/981

They were spurious and were happening every time the `onDisconnect` handler was invoked after `close` was invoked. That's a lot of the time! Better to have `close()` simply close the socket and have `onDisconnect` emit the relevant close type. Easier to understand now, too, I think? Tested manually with Simple SOCKS and `grunt test`.

Also include the TCP session ID in `SocksToRtc` and `RtcToNet` debugging so we can trace the whole lifecycle of a `tcp.ts` object (logging still a bit of a mess in there but getting better and better as we tackle these issues).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/204)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-networking",
   "description": "uProxy's networking library: SOCKS5 over WebRTC",
-  "version": "5.2.4",
+  "version": "5.2.5",
   "repository": {
     "type": "git",
     "url": "https://github.com/uProxy/uproxy-networking"
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -467,8 +467,8 @@ module RtcToNet {
     public longId = () : string => {
       var s = 'session ' + this.channelLabel();
       if (this.tcpConnection_) {
-        s += ' (TCP ' + (this.tcpConnection_.isClosed() ?
-            'closed' : 'open') + ')';
+        s += ' (socket ' + this.tcpConnection_.connectionId + ' ' +
+            (this.tcpConnection_.isClosed() ? 'closed' : 'open') + ')';
       }
       return s;
     }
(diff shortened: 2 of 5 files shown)


trevj:
@iislucas I think you're most familiar with tcp.ts.

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/204)

Reviewed files:
- package.json @ r2
- src/tcp/tcp.d.ts @ r1
- src/tcp/tcp.ts @ r1

---

One small idea. This is good. I like the comment fixes too :+1:

---

<sup>**[src/tcp/tcp.ts, line 370 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/204#-JizCZK65jaWrXkPII95)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/660bb2eba6e1209c7f20ce4c7c0b88326b580e64/src/tcp/tcp.ts#L370)):</sup>
Maybe also clear the queue? `this.dataToSocketQueue.clear();`

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/204)

---

Oh yeah, make sure this has been tested & passes the appropriate integration test. :)

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/204)

---

Many thanks.

Unit and integration tests passing and tested manually by browsing in Firefox and Chrome.

---

<sup>**[src/tcp/tcp.ts, line 370 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/204#-JizCZK65jaWrXkPII95)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/660bb2eba6e1209c7f20ce4c7c0b88326b580e64/src/tcp/tcp.ts#L370)):</sup>
Yes. I think that's a good idea. Done.

---

PR #206, opened 2015-02-27
Flattened JSON output and stringify early.

mollyling:
Makes it easier to parse from R and probably other stat tools.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/206)

### DIFF
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -25,6 +25,10 @@ module RtcToNet {
 
   export interface SessionSnapshot {
     name :string;
+    // Time in seconds, with fractional parts, of when the snapshot
+    // was taken.  Epoch is start of this web-worker.  This is the
+    // result of calling performance.now() -
+    // https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
     timestamp: number;
     channel_sent: number;
     channel_received: number;
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -23,28 +23,18 @@ module RtcToNet {
     allowNonUnicast :boolean;
   }
 
-  export interface HandlerQueueSnapshot {
-    size :number;
-    handling :boolean;
-  }
-
-  export interface SocketSnapshot {
-    sent :number;
-    received :number;
-    queue :HandlerQueueSnapshot;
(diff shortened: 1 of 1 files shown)


src/rtc-to-net/rtc-to-net.ts
```diff
@@ -23,28 +23,18 @@ module RtcToNet {
...
-    received :number;
-    buffered :number;
-    queue :HandlerQueueSnapshot;
-  }
-
   export interface SessionSnapshot {
     name :string;
-    channel :DataChannelSnapshot;
-    socket :SocketSnapshot;
+    timestamp: number;
```
trevj, line 32: Short comment here.

This is when the log entry was generated vs. the timestamp included with the log line?

src/rtc-to-net/rtc-to-net.ts
```diff
@@ -175,7 +165,7 @@ module RtcToNet {
       });
       var writeSnapshot = () => {
         this.getSnapshot().then((snapshot:RtcToNetSnapshot) => {
-          log.info('snapshot: %1', [JSON.stringify(snapshot)]);
+          log.info('snapshot: ' + JSON.stringify(snapshot));
```
trevj: Curious why you changed this -- the `%X` placeholders are intended to be faster, by avoiding the string concatenation.

trevj:
A couple of tiny things, otherwise :+1: and happy to see this evolving!

src/rtc-to-net/rtc-to-net.ts
mollyling: Two reasons:
- First, logging ain't working right, so I want to simplify what we're doing with logging.  Hopefully it'll eventually help
- Let's just construct and keep the string, and not retain the entire JS object.

src/rtc-to-net/rtc-to-net.ts
mollyling, line 32: Comment added.

src/rtc-to-net/rtc-to-net.ts
trevj: I don't see how it'll eventually help. FYI, @jpevarnek made some recent improvements and now you could also just do:

``` typescript
 log.info('snapshot: %1', snapshot);
```

Or even just:

``` typescript
 log.info(snapshot);
```

PR #207, opened 2015-02-27
Better request-size options in benchmark.

mollyling:
Adds in 128k, 256k, and 512k sizes for the generated data directory, and a -R option in the benchmark script to use them.  One small error fix (I don't know why the library gives me body but not body.length sometimes, but it doesn't matter...).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/207)

### DIFF
--- a/bin/make-data.sh
+++ b/bin/make-data.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 mkdir data
 pushd data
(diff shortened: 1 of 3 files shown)


trevj:
:+1:

trevj:
ping!

Is this change still relevant?

mollyling:
Ah, yes relevant.  Shippable just timed out on starting chrome.

PR #208, opened 2015-03-02
eliminate uncaught promise warnings and fix 

trevj:
Two main things in here:
- eliminate uncaught promise warnings in Chrome (namely `onceReady`)
- refactor `SocksToRtc.doRequestHandshake`, fixing an issue where it ignored unsuccessful connect messages from `RtcToNet`

I believe this addresses this issue:
https://github.com/uProxy/uproxy/issues/952

I also filed a couple of issues we should also look into:
1. https://github.com/uProxy/uproxy/issues/1032
2. https://github.com/uProxy/uproxy/issues/1030

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/208)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "uproxy-networking",
   "description": "uProxy's networking library: SOCKS5 over WebRTC",
-  "version": "5.3.0",
+  "version": "6.0.0",
   "repository": {
     "type": "git",
     "url": "https://github.com/uProxy/uproxy-networking"
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -339,9 +339,8 @@ module RtcToNet {
           }
           throw e;
         });
-      this.onceReady.then(this.linkSocketAndChannel_);
 
-      this.onceReady.catch(this.fulfillStopping_);
+      this.onceReady.then(this.linkSocketAndChannel_, this.fulfillStopping_);
 
       this.dataChannel_.onceClosed
       .then(() => {
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -218,8 +218,8 @@ module RtcToNet {
             channelLabel, Object.keys(this.sessions_).length]);
         };
       session.onceStopped().then(discard, (e:Error) => {
-        log.info('discarded session %1 (%2 remaining)', [
-            channelLabel, Object.keys(this.sessions_).length]);
+        log.error('session %1 terminated with error: %2', [
+            tag, e.message]);
         discard();
       });
     }
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -219,7 +219,7 @@ module RtcToNet {
         };
       session.onceStopped().then(discard, (e:Error) => {
         log.error('session %1 terminated with error: %2', [
-            tag, e.message]);
+            channelLabel, e.message]);
         discard();
       });
     }
--- a/src/socks-to-rtc/socks-to-rtc.d.ts
+++ b/src/socks-to-rtc/socks-to-rtc.d.ts
@@ -40,10 +40,10 @@ declare module SocksToRtc {
         dataChannel:WebRtc.DataChannel,
         bytesSentToPeer:Handler.Queue<number,void>,
         bytesReceivedFromPeer:Handler.Queue<number,void>)
-        => Promise<Net.Endpoint>;
+        => Promise<void>;
     public stop :() => Promise<void>;
     public tcpConnection :Tcp.Connection;
-    public onceReady :Promise<Net.Endpoint>;
+    public onceReady :Promise<void>;
     public onceStopped :Promise<void>;
     public longId :() => string;
     public channelLabel :() => string;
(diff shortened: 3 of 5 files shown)


iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208)

Reviewed files:
- package.json @ r1
- src/socks-to-rtc/socks-to-rtc.d.ts @ r1
- src/socks-to-rtc/socks-to-rtc.spec.ts @ r1

---

Remarks:  
- I think the old separation of `receiveResponseFromPeer_` from `doRequestHandshake_` might have improved readability. 
- Love the issues you observed and filed. Good thinking. 
- I think we want some way to tell a consumer of socks-rtc about failures, maybe by having `onceStopped` contain an enum value that expresses the reason it stopped? e.g. auth failure, normal close, etc. (filed here: https://github.com/uProxy/uproxy/issues/1037)

Good stuff; merge it (after you add a note on how you tested it, and confirm it works)  :+1:

---

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208)

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 451 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208#-JjVbmqcPhex7ylzvBP2)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/e75b0cf681293ced6262c21753e5840c2633df84/src/socks-to-rtc/socks-to-rtc.ts#L451)):</sup>
Minor code style issue: I try to reserve Errors for cases where the system is really malfunctioning.  There are lots of cases where you can get a non-SUCCEEDED error code when the system is functioning as intended, such as if the user types a local network address into the URL bar.

If you want all non-success response codes to trigger a rejection, it might be nicer to `return Promise.reject(...)`here.

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208)

---

Made a change in response to @bemasc's comments.

I've tested with `grunt test` and manually running these commands against Simple SOCKS:

``` bash
curl -x socks5h://localhost:9999 www.example.com
```

``` bash
curl -x socks5h://localhost:9999 www.exampdfgiuhfdsiguhfdguihdfiughle.com
```

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 451 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208#-JjVbmqcPhex7ylzvBP2)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/e75b0cf681293ced6262c21753e5840c2633df84/src/socks-to-rtc/socks-to-rtc.ts#L451)):</sup>
So, AFAIK, `throw`-ing  is treated identically to returning a rejected promise. If that's not the case, please let know!

Regarding throwing errors, I kinda agree. In this case, I think it  might be cleaner to have this function return a `Socks.Response` and have the caller inspect the `reply` type. This also delegates logging to the caller, which seems better.

PTAL.

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208)

---

Oh, @bemasc you made me realise that we should not be printing warnings every time a connection fails. Usually, as you say, it's because the user mis-typed the URL. I'd prefer treat that in a separate pull request:
https://github.com/uProxy/uproxy/issues/1039

---

trevj:
Fixed tests. PTAL!

bemasc:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208)

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 318 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208#-JjX9C7OK8TIhVRq-WFU)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/63a285a66c350c8d1b78e69c305fcb52550e88a6/src/socks-to-rtc/socks-to-rtc.ts#L318)):</sup>
Yes, the effect of throwing an Error is the same as rejecting the promise, but I think that an explicit reject better conveys that "this operation is returning a value indicating failure", whereas throwing an Error suggests "this application has encountered a bug or unexpected system failure".

---

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208)

Reviewed files:
- src/socks-to-rtc/socks-to-rtc.spec.ts @ r3
- src/socks-to-rtc/socks-to-rtc.ts @ r2

---

I agree with Ben's nit on throwing Error vs explicit reject promise. Otherwise :+1: from me. 

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208)

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 458 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208#-JjVbmqcPhex7ylzvBP2)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/e75b0cf681293ced6262c21753e5840c2633df84/src/socks-to-rtc/socks-to-rtc.ts#L458)):</sup>
So, aren't you supposed to reject promises with an `Error`?

---

trevj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/208)

---

So, after discussing in person, we're going to submit this and think a little more about how to distinguish success, (recoverable) exceptions, and (unrecoverable exceptions), a.k.a. errors.

---

PR #210, opened 2015-03-06
Mling integration tcp floodtests

mollyling:
Requires that https://github.com/uProxy/uproxy-lib/pull/137 is merged, to get the getStats() api from the queue handler.  This is on top of iislucas-requireify.

Run against the tcp flood server on its default port (1224) in the single-flood case, and run the server twice on the default (1224) and default+1 (1225) to let these test run against it.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/210)

### DIFF
--- a/src/integration-tests/tcp/freedom-module.ts
+++ b/src/integration-tests/tcp/freedom-module.ts
@@ -54,6 +54,8 @@ parentModule.on('floodtwoclient', (data) => {
   var second_total = 0.0;
   var first_start = 0;
   var second_start = 0;
+  var first_end = 0;
+  var second_end = 0;
   if (data) {
     if (data["firstport"]) {
       firstport = data.firstport;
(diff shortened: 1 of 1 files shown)


src/integration-tests/tcp/freedom-module.ts
```diff
@@ -46,6 +46,113 @@ parentModule.on('listen', () => {
   });
 });
 
+parentModule.on('floodtwoclient', (data) => {
```
iislucas: Add a comment with the command line arguments you might want to use to run this; e.g. using your server, nc or such.

iislucas:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/210)

---

Excellent! Trevor is going to like these a lot I think :) 

A few questions/name suggestions. 

---

<sup>**[src/integration-tests/tcp/freedom-module.ts, line 49 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/210#-JjlPyZhYgZi_lXRlic3)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/68bad08ee29ffe9b44764b3c8465f8bd3af297e9/src/integration-tests/tcp/freedom-module.ts#L49)):</sup>
Add a comment with the command line arguments you might want to use to run this; e.g. using your server, nc or such.

---

<sup>**[src/integration-tests/tcp/freedom-module.ts, line 67 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/210#-JjlQ1hYwAXINte7fHt-)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/68bad08ee29ffe9b44764b3c8465f8bd3af297e9/src/integration-tests/tcp/freedom-module.ts#L67)):</sup>
Before doing this, I think you should set the consoleFilter to not send stuff to the browser console log (only store it in the buffer), to do that you can use: 

```
loggingController.setConsoleFilter(['*:W']);
loggingController.setBufferedLogFilter(['*:D']);
```

Then to get logs back at the end of this test...

```
var logs = loggingController.getLogs(); 
```

Looking at the logging controller, it's pretty clear that the API needs to fixes: ability to get the current console filters; ability to clear all console filters; ability to disable console or buffered logs independently. Lets not fight around that stuff - when we need it, lets add it. 

---

<sup>**[src/integration-tests/tcp/freedom-module.ts, line 70 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/210#-JjlSOSK51dEAWZ18uCW)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/68bad08ee29ffe9b44764b3c8465f8bd3af297e9/src/integration-tests/tcp/freedom-module.ts#L70)):</sup>
Call this `logStatsLoop` instead of `fun` ? 

---

<sup>**[src/integration-tests/tcp/freedom-module.ts, line 98 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/210#-JjlSoOb3Jpuk-S5HvG3)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/68bad08ee29ffe9b44764b3c8465f8bd3af297e9/src/integration-tests/tcp/freedom-module.ts#L98)):</sup>
 I think we should also record the time-point before the tcp connection starts; tcp can take a while to start, and because of IPC things, you can actually start getting data before onceConnected completes :/  (this can be kind of fixed, and Ben and I have a plan for it...)

---

<sup>**[src/integration-tests/tcp/freedom-module.ts, line 121 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/210#-JjlTtBkIyZMhEhGfYeB)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/68bad08ee29ffe9b44764b3c8465f8bd3af297e9/src/integration-tests/tcp/freedom-module.ts#L121)):</sup>
name fun better... no, not literally... ;-)

---

<sup>**[src/integration-tests/tcp/freedom-module.ts, line 125 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/210#-JjlTfq75zvWIb7y46OT)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/68bad08ee29ffe9b44764b3c8465f8bd3af297e9/src/integration-tests/tcp/freedom-module.ts#L125)):</sup>
Maybe better to name these in the same way as above? 

---

<sup>**[src/integration-tests/tcp/freedom-module.ts, line 140 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/210#-JjlUDXV7QVn9DRmm9Pq)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/68bad08ee29ffe9b44764b3c8465f8bd3af297e9/src/integration-tests/tcp/freedom-module.ts#L140)):</sup>
How are these logs distinguished from the ones in the 1 second timeout loop? Don't they want to be? 

---

trevj:
OK, this has been out a while. Sorry about that. You're probably best rebasing and basing this off the current dev branch. Thanks for adding these!

Two questions:
- these tests don't have pass/fail conditions so I wonder 
- I think you were invoking these via the console? That's a neat trick but I think they're better placed in some other files, a little like the `integrationSlow` target: https://github.com/uProxy/uproxy-networking/blob/dev/Gruntfile.coffee#L151

(sorry I know this is a pain but hopefully it's not too annoying to copy pasta another target similar to `integrationSlow`)

iislucas:
(Assigned to @trevj ; what do we want to do with this?)

trevj:
Guess we can close this!

PR #214, opened 2015-03-19
Pause and resume around the initial getInfo.

bemasc:
This should hopefully enable faster startup on fast downloads.

Tested by running the TCP integration tests (which pass).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/214)


iislucas:
LGTM. Please add how it was tested to the description above.

PR #215, opened 2015-03-19
Fix paths in Gruntfile for socks-echo integration tests.

bemasc:
This does not cause the tests to run (they are still
blocked by an error related to require.js's use of string
evaluation) but at least it loads the right files.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/215)


iislucas:
Oops, we both worked on this. :)   I think I have these fixed in a local branch with the new grunt-jasmine-chromeapp... :)

bemasc:
OK, no problem.  I just wanted to make sure that this didn't get forgotten.

PR #217, opened 2015-03-20
experiment: churn pipe flattening

trevj:
This eliminates one of the UDP sockets involved with churn. Now, churn binds to and sends and receives all messages from the probe connection's local UDP port. I believe this is safe but would be interested to hear what you think!

Rough performance testing on my machine showed curl performance jumping from 100-150k/sec to ~200k/sec.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/217)


bemasc:
I expect that the speedup is due to removing the freedom module boundary, not due to reducing the number of sockets.

I'm not convinced that reducing the number of sockets is a good idea.  I think having independent sockets is clearer than dispatching based on source address.  Also, to do port mirroring (which is approaching the front of my task queue) we're going to end up managing a plurality of sockets anyway.

So could we split those two changes?

willscott:
Do you have a sense of what else is happening here that is keeping performance at 200k versus the 750k you were able to obtain without use of the obfuscator? Is it unwrapping the buffers to look at the bytes that's the real slowdown here?

trevj:
@bemasc I agree any speedup is almost certainly the result of reduced message passing. However, I don't agree what we have now is simpler: it may be superficially more elegant to characterise a greater number of sockets as "independent" but I feel the code (and resultant diagrams, if I were to attempt to draw this) speak for themselves. Anyway, this is not ready to submit yet -- just wanted to solicit opinions, thanks! BTW, it seems to me from reading the design doc that port mirroring also requires examination of packets' source addresses?

@willscott  I don't have a good sense of what's causing the slowdown but it's not examining the bytes: even if churn forwards the bytes opaquely without any transformation, I see no noticeable improvement. I have observed very strange behaviour with message handling when churn is under high load but it's not clear to me where the problem lies (VM, freedom, churn, Chrome sockets, etc.). @mollyling will be trying to get to the bottom of this over the next week or two.

iislucas:
Lets do a campfire discussion on the possible churn deigns; it's been a long time.

PR #220, opened 2015-04-01
Pause the incoming socket when the downstream data channel overflows.

bemasc:
This change was tested using simple socks and the flood server.

Note that for testing, simple socks must be configured not to use
CHURN, and debug logging must be disabled.

This resolves https://github.com/uProxy/uproxy/issues/949.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/220)

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "socks5-http-client": "^0.1.6",
     "tsd": "^0.5.7",
     "typescript": "^1.4.1",
-    "uproxy-lib": "^20.0.0",
+    "uproxy-lib": "^20.1.0",
     "utransformers": "^0.2.1",
     "wup": "^1.0.0",
     "yargs": "^3.0.4"
--- a/src/net/tcp.ts
+++ b/src/net/tcp.ts
@@ -270,7 +270,7 @@ export class Connection {
   private fulfillClosed_ :(reason:SocketCloseKind)=>void;
 
   // A TCP connection for a given socket.
-  constructor(connectionKind:Connection.Kind) {
+  constructor(connectionKind:Connection.Kind, private startPaused_?:boolean) {
     this.connectionId = 'N' + Connection.globalConnectionId_++;
 
     this.dataFromSocketQueue = new handler.Queue<ArrayBuffer,void>();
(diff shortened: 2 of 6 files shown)


iislucas:
Would be great to use @mollyling  performance benchmarks to get some graphs that show the difference this makes & how if fairs at different loads.

iislucas:
@mollyling Can you have a look at this while I'm out?

mollyling:
Sure

Cheers,
$ ls

bemasc:
@trevj PTAL

---

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/220)

trevj:
Two things:
1. don't forget to add the session ID to the new logging messages (I have a comment on this from an earlier round)
2. you still have a `.isClosed` check in SocksToRtc -- I think in general whatever you did for RtcToNet should be copy and paste-able there

Then, :+1:.

---

---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 604 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/220#-JmExdbWJWjCRqzC4S3d)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d61a6e05c11a29539a97084b97461945c54148b7/src/rtc-to-net/rtc-to-net.ts#L604)):</sup>
Let's add the ID, e.g.:

``` typescript
log.debug('%1: hit overflow, pausing', this.longId());
```

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/220)

bemasc:
Thanks for the catch!  I made SocksToRtc and RtcToNet parallel.

---

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/220)

bemasc:
---

<sup>**[src/rtc-to-net/rtc-to-net.ts, line 596 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/220#-JmExdbWJWjCRqzC4S3d)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/d61a6e05c11a29539a97084b97461945c54148b7/src/rtc-to-net/rtc-to-net.ts#L596)):</sup>
Done.

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/220)

PR #222, opened 2015-04-03
Fix installation instructions (and a few typos while we're at it).

matthild:
Following the documented steps in a freshly cloned repository resulted
in a build of `copypaste-socks-chromeapp` that showed simply an empty white
window.  After `bower install` and rebuilding, the
`copypaste-socks-chromeapp` is nonempty.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/222)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,8 @@ transform and restore the data being sent over the network.
 
 ## Building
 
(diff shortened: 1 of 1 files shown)


iislucas:
:+1:

PR #229, opened 2015-04-10
move to return-promise-on-close versions of uproxy-lib

trevj:
Just a couple of simple updates se we can make use of:
- https://github.com/uProxy/uproxy-lib/pull/150
- https://github.com/uProxy/uproxy-lib/pull/149

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/229)


bemasc:
:+1:

PR #233, opened 2015-04-14
Implement port mirroring.

bemasc:
This allows CHURN connections to peers who are behind symmetric NAT.

https://github.com/uProxy/uproxy/issues/608

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/233)

### DIFF
--- a/src/churn/churn.spec.ts
+++ b/src/churn/churn.spec.ts
@@ -105,14 +105,25 @@ describe("selectPublicAddress", function() {
     port: 56635
   };
 
+  var publicEndpoint :net.Endpoint = {
+    address: '18.19.20.21',
+    port: 10011
+  };
+
   var srflxCandidate = {
     candidate: 'a=candidate:129713316 2 udp 2122129151 ' +
       srflxEndpoint.address + ' ' + srflxEndpoint.port + ' typ srflx raddr ' +
       baseEndpoint.address + ' rport ' + baseEndpoint.port + ' generation 0'
   };
+
   var hostCandidate = {
     candidate: 'a=candidate:9097 1 udp 4175 ' + baseEndpoint.address + ' ' +
         baseEndpoint.port + ' typ host generation 0'
+  };
+
+  var publicHostCandidate = {
+    candidate: 'a=candidate:1234321 1 udp 4567 ' + publicEndpoint.address + ' ' +
+        publicEndpoint.port + ' typ host generation 0'
   };
 
   var relayCandidate = {
@@ -165,6 +176,17 @@ describe("selectPublicAddress", function() {
     expect(natPair).toEqual(correctNatPair);
 
     natPair = churn.selectPublicAddress([hostCandidate, relayCandidate]);
+    expect(natPair).toEqual(correctNatPair);
+  });
+
+  it('use public host if it is present', () => {
+    var correctNatPair = {
+      internal: publicEndpoint,
+      external: publicEndpoint
+    };
+
+    var natPair = churn.selectPublicAddress(
+        [hostCandidate, relayCandidate, srflxCandidate, publicHostCandidate]);
     expect(natPair).toEqual(correctNatPair);
   });
 });
--- a/src/churn/churn.ts
+++ b/src/churn/churn.ts
@@ -18,6 +18,7 @@
 
 import arraybuffers = require('../../../third_party/uproxy-lib/arraybuffers/arraybuffers');
 import churn_pipe_types = require('../churn-pipe/freedom-module.interface');
+import ChurnPipe = churn_pipe_types.freedom_ChurnPipe;
 import churn_types = require('./churn.types');
 import handler = require('../../../third_party/uproxy-lib/handler/queue');
 import ipaddr = require('ipaddr.js');
@@ -149,6 +150,28 @@ var log :logging.Log = new logging.Log('churn');
     throw new Error('no srflx or host candidate found');
   };
 
+  var endpointKey_ = (endpoint:net.Endpoint) : string => {
+    return endpoint.address + ':' + endpoint.port;
+  };
+
+  // Retry an async function with exponential backoff for up to 2 seconds
+  // before failing.
+  var retry_ = (func:() => Promise<void>, delay?:number) : Promise<void> => {
+    var delay = delay || 10;  // milliseconds
+    var maxDelay = 2000;  // milliseconds
+    return func().catch((err) => {
+      if (delay > maxDelay) {
+        log.error('Timeout reached while retrying');
+        return Promise.reject(err);
+      }
+      return new Promise<void>((F, R) => {
+        setTimeout(() => {
+          this.retry_(func, delay * 2).then(F, R);
+        }, delay);
+      });
+    });
+  }
+
   /**
    * A uproxypeerconnection-like Freedom module which establishes obfuscated
    * connections.
@@ -206,13 +229,17 @@ var log :logging.Log = new logging.Log('churn');
       this.haveRemoteEndpoint_ = F;
     });
 
-    // Fulfills once we've successfully allocated the forwarding socket.
+    // Fulfills once we've successfully allocated the mirror pipe representing the
+    // remote peer's signalled transport address.
     // At that point, we can inject its address into candidate messages destined
     // for the local RTCPeerConnection.
     private haveForwardingSocketEndpoint_ :(endpoint:net.Endpoint) => void;
     private onceHaveForwardingSocketEndpoint_ = new Promise((F, R) => {
       this.haveForwardingSocketEndpoint_ = F;
     });
+
+    // A map from remote transport addresses to local pipes that represent them.
+    private mirrorPipes_ : { [k: string]: ChurnPipe } = {};
 
     private static internalConnectionId_ = 0;
 
@@ -284,24 +311,43 @@ var log :logging.Log = new logging.Log('churn');
         if (message.type === signals.Type.CANDIDATE) {
           this.probeCandidates_.push(message.candidate);
         } else if (message.type === signals.Type.NO_MORE_CANDIDATES) {
-          this.probeConnection_.close();
-          this.probingComplete_(selectPublicAddress(this.probeCandidates_));
+          this.probeConnection_.close().then(() => {
+            this.probingComplete_(selectPublicAddress(this.probeCandidates_));
+          });
         }
       });
       this.probeConnection_.negotiateConnection();
     }
 
-    // Establishes the two pipes required to sustain the obfuscated
-    // connection:
-    //  - a non-obfuscated, local only, between WebRTC and a new,
-    //    automatically allocated, port
-    //  - remote, obfuscated, port
-    private configurePipes_ = (
+    // Add and return a local pipe that represents a specific remote address
+    // for both send and receive.
+    private addLocalPipe_ = (
         webRtcEndpoint:net.Endpoint,
         remoteEndpoint:net.Endpoint,
-        natEndpoints:NatPair) : void => {
+        publicPipe:ChurnPipe)
+        : Promise<ChurnPipe> => {
+      var key = endpointKey_(remoteEndpoint);
+      if (this.mirrorPipes_[key]) {
+        log.warn('%1: Got redundant call to add local pipe for %2',
+            this.peerName,
+            key);
+        // Return the pipe, but wait until it's ready.
+        return this.mirrorPipes_[key].getLocalEndpoint().then(
+            (ignored:net.Endpoint) => {
+          return this.mirrorPipes_[key];
+        });
+      }
+
       var localPipe = freedom['churnPipe']();
-      localPipe.bind(
+      this.mirrorPipes_[key] = localPipe;
+
+      // Packets received by this pipe should be obfuscated and forwarded
+      // to the corresponding remote endpoint.
+      localPipe.on('message', (m:churn_pipe_types.Message) => {
+        publicPipe.sendTo(m.data, remoteEndpoint);
+      });
+
+      return localPipe.bind(
           '127.0.0.1',
           0,
           webRtcEndpoint.address,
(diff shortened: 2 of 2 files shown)


trevj:
Two questions, before going much further:
1. Have you been able to run this for real?
2. It looks like it might be eliminate the distinction between the "original" local pipe and mirror pipes, if all local pipes were created on-demand. Does that make sense? Did you consider this?

---

<sup>**[src/churn/churn.ts, line 21 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuG1LfYFbFOG1xz0g)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L21)):</sup>
The convention seems to be lowercase, i.e. `churn_pipe`.

---

<sup>**[src/churn/churn.ts, line 153 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuI1bRk4nXSklWvMm)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L153)):</sup>
Short comment, e.g.:

``` js
// Returns a key for use with mirrorPipes_.
```

---

<sup>**[src/churn/churn.ts, line 153 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuKOx7Ibp8kcoGv3F)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L153)):</sup>
What do you think of renaming this to `getEndpointKey_` or `makeEndpointKey_`?

---

<sup>**[src/churn/churn.ts, line 159 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsvqUCWtqxLA5kaM7Z)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L159)):</sup>
See my note below on how I've worked around this issue previously.

---

<sup>**[src/churn/churn.ts, line 160 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuPduVJPZpYn7r1Rx)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L160)):</sup>
Suggestion: rename `delayMs`. This would eliminate the need for a comment.

---

<sup>**[src/churn/churn.ts, line 161 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuWXMz37CaPvh_wPK)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L161)):</sup>
This can be a static constant.

---

<sup>**[src/churn/churn.ts, line 163 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuYLy1UWiKTfSOKsN)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L163)):</sup>
This doesn't look quite right: shouldn't this be a precondition-like statement at the top of `rety_`? This way, `func` will be called (once) after 2s.

---

<sup>**[src/churn/churn.ts, line 164 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuafkVwcyN2raZhYz)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L164)):</sup>
Don't log-and-throw. Better to allow the caller decide to log, when possible.

---

<sup>**[src/churn/churn.ts, line 396 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsvXXdVKoLuoaAQ9nW)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L396)):</sup>
I ran into the same problem when I drafted my "churn pipe flattening" experiment and was able to workaround it with a "throwaway port":
https://github.com/uProxy/uproxy-networking/compare/trevj-churn-pipe-flattening#diff-63e88a30f0d785092bff0b224c9704a3R297

Slightly wasteful and weird but is certainly simpler than the `retry_` function.

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233)

bemasc:
<sup>**[src/churn/churn.ts, line 21 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuG1LfYFbFOG1xz0g)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L21)):</sup>
I think the convention is for namespaces to be lowercase, but for class instance types to be capitalized.  I moved this import (which is really an "alias") down to a more comprehensible location.

---

<sup>**[src/churn/churn.ts, line 153 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuI1bRk4nXSklWvMm)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L153)):</sup>
Done.

---

<sup>**[src/churn/churn.ts, line 153 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuKOx7Ibp8kcoGv3F)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L153)):</sup>
Done.  Changed to `makeEndpointKey_`

---

<sup>**[src/churn/churn.ts, line 160 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuPduVJPZpYn7r1Rx)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L160)):</sup>
Done.

---

<sup>**[src/churn/churn.ts, line 161 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuWXMz37CaPvh_wPK)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L161)):</sup>
Done.

---

<sup>**[src/churn/churn.ts, line 163 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuYLy1UWiKTfSOKsN)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L163)):</sup>
I think it has to be here, so that we can pass on the same rejection produced by `func`.

---

<sup>**[src/churn/churn.ts, line 164 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuafkVwcyN2raZhYz)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L164)):</sup>
Done.

---

<sup>**[src/churn/churn.ts, line 396 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsvXXdVKoLuoaAQ9nW)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L396)):</sup>
You're right, it's the same problem.  I think your code there essentially uses the socket creation to introduce a delay, which in testing appears to be longer than the time required for WebRTC to unbind the socket.

If we're going to introduce a delay to resolve the race, I'd rather it be an explicit delay, as in this retry function.

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233)

bemasc:
1. Only simple-socks
2. There isn't really any distinction, except that the original local pipe is created based on an incoming candidate, rather than an incoming packet.  This is necessary in order to send outgoing traffic to the remote address before receiving incoming traffic.  Ultimately, I think the rule will be "a mirror for every candidate and for every unexpected source port".

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233)

trevj:
1. I think you need to try this outside of Simple SOCKS!
2. Ah...so it wouldn't work to delay creation of the "local pipe" until we receive a packet from it?

---

<sup>**[src/churn/churn.ts, line 161 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuPduVJPZpYn7r1Rx)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L161)):</sup>
You don't really need the comment now (though it's fine, of course, if you still like it).

---

<sup>**[src/churn/churn.ts, line 163 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuYLy1UWiKTfSOKsN)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L163)):</sup>
Ah...I see you are multiplying `delayMs` before rejecting. So this works. Sorry!

FWIW, this what I was thinking:

``` typescript
var retry_ = (func:() => Promise<void>, delayMs?:number) : Promise<void> => {
  delayMs = delayMs || 10;
  if (delayMs > 2000) {
    return Promise.reject(err);
  }
  return func().catch((err) => {
    return new Promise<void>((F, R) => {
      setTimeout(() => {
        this.retry_(func, delayMs * 2).then(F, R);
      }, delayMs);
    });
  });
}
```

---

<sup>**[src/churn/churn.ts, line 396 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsvXXdVKoLuoaAQ9nW)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L396)):</sup>
I can't disagree :-)

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233)

bemasc:
OK, I tried this for real, using the copypaste CHURN chat chromeapp, between my workstation (symmetric NAT) and home server box (public IP).  Getting it working required some additional changes to ensure that the server box actually picked a non-NAT candidate.

---

<sup>**[src/churn/churn.ts, line 161 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuPduVJPZpYn7r1Rx)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L161)):</sup>
Done.

---

<sup>**[src/churn/churn.ts, line 163 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuYLy1UWiKTfSOKsN)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L163)):</sup>
Yeah, that is nicer but won't work because `err` is not defined.

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233)

trevj:
Awesome, thanks for testing it for real. Very glad to trapped that issue, which I can't imagine was much fun.

Just a couple of suggestions related to comments, then :+1:.

---

<sup>**[src/churn/churn.ts, line 86 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-Jo5SVfyy20aQc4MqFAj)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ace9b7ec60b0bdd0a3a2afe04ea51ba709880506/src/churn/churn.ts#L86)):</sup>
Let's quickly describe the heuristic here, e.g. favours server reflexive over host candidates, unless a host candidate appears to be a public IP.

---

<sup>**[src/churn/churn.ts, line 86 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-Jo5ToK58LZfOx_3qUTo)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ace9b7ec60b0bdd0a3a2afe04ea51ba709880506/src/churn/churn.ts#L86)):</sup>
You mentioned replacing (or augmenting) this heuristic with a "try every candidate" approach -- does this issue cover that?:
https://github.com/uProxy/uproxy/issues/1107

---

<sup>**[src/churn/churn.ts, line 171 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-JmsuYLy1UWiKTfSOKsN)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/65abd6ac38a393f01aa9f6f2c7e9bbd047daa8bc/src/churn/churn.ts#L171)):</sup>
Oops. I know code. Sorry :-)

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233)

bemasc:
<sup>**[src/churn/churn.ts, line 86 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-Jo5SVfyy20aQc4MqFAj)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ace9b7ec60b0bdd0a3a2afe04ea51ba709880506/src/churn/churn.ts#L86)):</sup>
Done.

---

<sup>**[src/churn/churn.ts, line 86 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233#-Jo5ToK58LZfOx_3qUTo)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/ace9b7ec60b0bdd0a3a2afe04ea51ba709880506/src/churn/churn.ts#L86)):</sup>
I think that's probably close enough.

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/233)

PR #235, opened 2015-04-14
fix path in copypaste sample freedom manifest

trevj:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/235)


iislucas:
:+1:

PR #236, opened 2015-04-15
Make dist files be more sensible; make integration test to not be interactive

iislucas:
TESTED: 
- grunt dist

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/236)


trevj:
:+1:

PR #237, opened 2015-04-15
Gruntfile tweak

iislucas:
Fix to avoid grunt-jasmine-chromeapp race condition (it exits before it has really finished, which resulted in copy:dist rule failing). This is better anyway as it gives a dist directory independently of integration tests passing/failing (which is occasionally useful)

TESTED: 
- grunt dist

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/237)


iislucas:
got LGTM from @trevj

PR #245, opened 2015-04-19
fix bad style for freedom module creation

iislucas:
This pull request fixes the object creation style for logging controllers. 

Context: Freedom module creation is an operation that requires cleanup. So it's better style to name the object that gets created so that you have the ability to clean it up, and so that you don't end up creating multiple stubs in memory (e.g. when you want to add more/different logging filters). 

This pull request is also helpful as it lets you find the appropriate object in the webworker in case you want to interactively set/change filter levels. 

TESTED: 
grunt dist

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/245)

### DIFF
--- a/src/samples/simple-turn-chromeapp/freedom-module.ts
+++ b/src/samples/simple-turn-chromeapp/freedom-module.ts
@@ -10,8 +10,9 @@ import net = require('../../net/net.types');
 import logging = require('../../../../third_party/uproxy-lib/logging/logging');
 import loggingTypes = require('../../../../third_party/uproxy-lib/loggingprovider/loggingprovider.types');
 
-freedom['loggingcontroller']().setDefaultFilter(loggingTypes.Destination.console,
-                                                loggingTypes.Level.info);
+var loggingController = freedom['loggingcontroller']();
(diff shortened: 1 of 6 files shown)


trevj:
:+1:

src/samples/simple-turn-chromeapp/freedom-module.ts
```diff
@@ -10,8 +10,9 @@ import net = require('../../net/net.types');
 import logging = require('../../../../third_party/uproxy-lib/logging/logging');
 import loggingTypes = require('../../../../third_party/uproxy-lib/loggingprovider/loggingprovider.types');
 
-freedom['loggingcontroller']().setDefaultFilter(loggingTypes.Destination.console,
-                                                loggingTypes.Level.info);
+var loggingController = freedom['loggingcontroller']();
+loggingController = setDefaultFilter(loggingTypes.Destination.console,
```
salomegeo: = ?

iislucas: Ah, yes, fixed. :)

PR #246, opened 2015-04-20
remove last remnants of nextTick hack

trevj:
This removes the last few remaining uses of `nextTick`. Testing locally, this just doesn't seem to be necessary anymore, even in the `dev` branch. I'm able to leave an ISO downloading in the background while browsing just fine.

I'm not sure exactly when this hack became obsolete - although TCP backpressure is the most likely candidate - but since we were hazy about why it ever worked in the first place, I'm more than happy to remove it -- especially since the raw download speed on my machine with this branch is >2x that of dev (~1900Kb/sec vs. ~850Kb/sec).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/246)

### DIFF
--- a/src/rtc-to-net/rtc-to-net.ts
+++ b/src/rtc-to-net/rtc-to-net.ts
@@ -280,10 +280,6 @@ import logging = require('../../../third_party/uproxy-lib/logging/logging');
   export class Session {
     private tcpConnection_ :tcp.Connection;
 
-    // TODO: There's no equivalent of datachannel.isClosed():
-    //         https://github.com/uProxy/uproxy/issues/1075
-    private isChannelClosed_ :boolean = false;
-
     // Fulfills once a connection has been established with the remote peer.
     // Rejects if a connection cannot be made for any reason.
     public onceReady :Promise<void>;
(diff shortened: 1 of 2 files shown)


bemasc:
<sup>**[src/rtc-to-net/rtc-to-net.ts, line 354 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/246#-JnNHcksf6VMrGo2vtuS)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/64b4c67d3c6e8265e13829ac1d49aba383672fc0/src/rtc-to-net/rtc-to-net.ts#L354)):</sup>
Are you sure it's safe to remove this check?  It seems like this queue is guaranteed to be empty after linkSocketAndChannel, but what if onceClosed runs before that?

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/246)

trevj:
<sup>**[src/rtc-to-net/rtc-to-net.ts, line 354 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/246#-JnNHcksf6VMrGo2vtuS)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/64b4c67d3c6e8265e13829ac1d49aba383672fc0/src/rtc-to-net/rtc-to-net.ts#L354)):</sup>
Good catch. As discussed, this logic isn't impeccable generally but should, in this case, be harmless. Let's log a WARNing when this happens.

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/246)

PR #247, opened 2015-04-20
add SocketCloseKind to TCP server onceClosed, with cleanup

trevj:
While investigating seemingly-spontaneous Simple SOCKS shutdowns today, I had reason to suspect the TCP server was the real culprit.

It's currently poorly instrumented, so this is an effort to address that. Aside from the formatting cleanup, this adds `SocketCloseKind` to `onceClosed`. I think that's worth doing but am curious to hear your thoughts.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/247)

### DIFF
--- a/src/net/tcp.ts
+++ b/src/net/tcp.ts
@@ -65,6 +65,12 @@ function destroyFreedomSocket_(socket:freedom_TcpSocket.Socket) : Promise<void> 
 // Promise and handler queue-based TCP server with freedomjs sockets.
 // TODO: protection against multiple calls to methods such as listen
 export class Server {
+  // Count of sockets created to date.
+  private static numCreations_ :number = 0;
+
+  // Unique ID, for logging purposes.
+  private id_ :string;
+
   private socket_ :freedom_TcpSocket.Socket;
 
   // Active connections to the server.
@@ -94,6 +100,8 @@ export class Server {
 
   constructor(private endpoint_ :net.Endpoint,
       private maxConnections_ :number = DEFAULT_MAX_CONNECTIONS) {
+    this.id_ = 'S' + (Server.numCreations_++);
+
     this.onceListening_.then((unused:any) => {
       this.isListening_ = true;
     }, (e:Error) => {
@@ -111,7 +119,7 @@ export class Server {
 
   // Invoked when the socket terminates.
   private onDisconnectHandler_ = (info:freedom_TcpSocket.DisconnectInfo) : void => {
-    log.debug('disconnected: %1', JSON.stringify(info));
+    log.debug('%1: disconnected: %2', this.id_, JSON.stringify(info));
     if (info.errcode === 'SUCCESS') {
       this.fulfillShutdown_(SocketCloseKind.WE_CLOSED_IT);
     } else {
@@ -141,11 +149,12 @@ export class Server {
   // Invoked each time a new connection is established with the server.
   private onConnectionHandler_ = (
       acceptValue:freedom_TcpSocket.ConnectInfo) : void => {
-    log.debug('new connection');
+    log.debug('%1: new connection', this.id_);
     var socketId = acceptValue.socket;
 
     if (this.connectionsCount() >= this.maxConnections_) {
-      log.warn('hit maximum connections count, dropping new connection');
+      log.warn('%1: hit maximum connections count, dropping new connection',
+          this.id_);
       destroyFreedomSocket_(freedom['core.tcpsocket'](socketId));
       return;
     }
@@ -157,10 +166,12 @@ export class Server {
 
     var discard = () => {
       delete this.connections_[socketId];
-      log.debug('discarded connection (%1 remaining)', this.connectionsCount());
+      log.debug('%1: discarded connection (%2 remaining)',
+          this.id_, this.connectionsCount());
     };
     connection.onceClosed.then(discard, (e:Error) => {
-      log.error('connection %1 rejected on close: %2', socketId, e.message);
+      log.error('%1: connection %2 rejected on close: %3',
+          this.id_, socketId, e.message);
       discard();
     });
 
@@ -170,7 +181,7 @@ export class Server {
   // Closes the server socket then closes all active connections.
   // Equivalent to calling stopListening followed by closeAll.
   public shutdown = () : Promise<void> => {
-    log.debug('shutdown');
+    log.debug('%1: shutdown', this.id_);
     // This order is important: make sure no new connections happen while
     // we're trying to close all the connections.
     return this.stopListening().then(this.closeAll);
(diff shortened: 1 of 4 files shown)


trevj:
@iislucas you might be interested, too

src/net/tcp.ts
```diff
@@ -51,198 +46,204 @@ export function endpointOfSocketInfo(info:freedom_TcpSocket.SocketInfo)
...
-  public onceShutdown = () : Promise<void> => { return this.onceShutdown_; }
-  // isShutdown can become true, but never false again. It gives synchronous
-  // access to test if |onceShutdown_| fulfilled.
-  private isShutdown_ :boolean = false;
-  public isShutdown = () : boolean => { return this.isShutdown_; }
+  private socket_ :freedom_TcpSocket.Socket;
+
+  // Active connections to the server.
+  // TODO: index by connectionId rather than socketID
+  private connections_:{[socketId:number] : Connection} = {};
```
bemasc: Space

trevj: Done.

trevj:
I also filed this issue:
https://github.com/uProxy/uproxy/issues/1316

iislucas:
LGTM

Makes me think it might be time to add that smart promise to uproxy lib - we really are using the same abstraction over and over again (Promise with a fulfill state + fulfill & reject functions). But that can happen later too.

---

Reviewed files:
- src/echo/tcp-echo-server.ts @ r1
- src/net/tcp.ts @ r2

---

<sup>**[src/net/tcp.ts, line 96 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247#-JnRFyINFLSu7gpRkcG8)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/7585c12883acc8cafdf3201b3b7c2a0c44b6201e/src/net/tcp.ts#L96)):</sup>
If you want, you can specify a default in the constructor like so: 

```
private maxConnections_ :number = DEFAULT_MAX_CONNECTIONS
```

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247)

iislucas:
(Note: make sure you say how this was tested)

bemasc:
<sup>**[src/net/tcp.ts, line 116 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247#-JnRNtRBY1UuYwLOJe-U)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/7585c12883acc8cafdf3201b3b7c2a0c44b6201e/src/net/tcp.ts#L116)):</sup>
Should all these log message be prepended with an object identifier, like in our usual style?  Maybe the endpoint, or even just the port, would do?

---

<sup>**[src/net/tcp.ts, line 193 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247#-JnRORbFKPU6aVOL4qJX)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/7585c12883acc8cafdf3201b3b7c2a0c44b6201e/src/net/tcp.ts#L193)):</sup>
I think I'd prefer `socketId` as a name

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247)

trevj:
The fulfill/rejection function pattern for sure, although in this case I only kept `isListening_` and `isShutdown_` to avoid changing the public interface too much because it's (still, sorry!) not clear to me that `isXxx_` is necessary, or a good idea.

I really find it interesting how this seemingly trivial class showcases several long-standing design questions:
1. the afore-mentioned "smart promises"
2. `doSomething` value-returning functions _plus_ `onceSomething` functions (what if `something` can be invoked many times on an object?)
3. public fields (not a question in my mind: I'm 100% against these)
4. interaction between promises, viz. `onceListening` and `onceShutdown` (should the second reject when the first fails? I say no)
5. `toString` methods -- I'm still not sure when these are invoked

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247)

trevj:
<sup>**[src/net/tcp.ts, line 96 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247#-JnRFyINFLSu7gpRkcG8)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/7585c12883acc8cafdf3201b3b7c2a0c44b6201e/src/net/tcp.ts#L96)):</sup>
Thanks!

---

<sup>**[src/net/tcp.ts, line 116 [r2]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247#-JnRNtRBY1UuYwLOJe-U)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/7585c12883acc8cafdf3201b3b7c2a0c44b6201e/src/net/tcp.ts#L116)):</sup>
It's no harm , and easy to do. Rather than use the endpoint - which may change over time - I used a unique ascending ID, like connections. We should think about better ways to do this in future, e.g. maintain a per-instance Logger each with its own name. I filed an issue to think about this:
https://github.com/uProxy/uproxy/issues/1350

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247)

bemasc:
:+1: 

---

<sup>**[src/net/tcp.ts, line 214 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247#-JnTUY3zEbygCEeyTk01)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/319d01e47489e7cc19265e179d3f865fc268190d/src/net/tcp.ts#L214)):</sup>
`socketId` here too

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247)

trevj:
Thanks for the reviews!

A few more things, then I'd like to submit:
- make `SocksToRtc` log the server socket close kind, so we can distinguish planned shutdowns from unplanned
- remove the `isListening` and `isShutdown` methods, which weren't being used anywhere and can come back at some later stage when we have "smart promises"
- tidy the `toString` method

Lastly, I bumped the minor version number figuring that going from Promise<void> to Promise<anything else> is equivalent to adding a return type to a method, i.e. does not break anything -- also, nothing outside of this repo uses `tcp.ts` directly so it's arguably not part of the public interface.

PTAL!

---

<sup>**[src/net/tcp.ts, line 214 [r3]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247#-JnTUY3zEbygCEeyTk01)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/319d01e47489e7cc19265e179d3f865fc268190d/src/net/tcp.ts#L214)):</sup>
Done.

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/247)

trevj:
I almost forgot -- tested by:
- running the echo server sample
- running copypaste and simple SOCKS
- inspecting the objects in the JavaScript debugger

bemasc:
:+1: (and make sure to check armoryonpark, if you haven't already!)

trevj:
Yep, armoryonpark works. Thanks!

PR #248, opened 2015-04-22
Remove obsolete SocksToRtc.sessions_ attribute

bemasc:
This will at least get rid of the error messages noted in https://github.com/uProxy/uproxy/issues/1346.  (Whether it will fix the behavioral failure is unknown.)

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/248)

### DIFF
--- a/src/socks-to-rtc/socks-to-rtc.ts
+++ b/src/socks-to-rtc/socks-to-rtc.ts
@@ -218,8 +218,8 @@ module SocksToRtc {
     private makeTcpToRtcSession_ = (tcpConnection:tcp.Connection) : void => {
       this.pool_.openDataChannel()
           .then((channel:peerconnection.DataChannel) => {
-        var tag = channel.getLabel();
-        log.info('associating channel %1 with new SOCKS client', tag);
+        var channelLabel = channel.getLabel();
+        log.info('associating channel %1 with new SOCKS client', channelLabel);
         var session = new Session();
         session.start(
             tcpConnection,
@@ -228,14 +228,14 @@ module SocksToRtc {
             this.bytesReceivedFromPeer_)
         .catch((e:Error) => {
           log.warn('session %1 failed to connect to remote endpoint: %2', [
-              tag, e.message]);
+              channelLabel, e.message]);
         });
 
         session.onceStopped.then(() => {
-          log.info('discarded session %1', tag);
+          log.info('discarded session %1', channelLabel);
         }, (e:Error) => {
           log.error('session %1 terminated with error: %2', [
-              tag, e.message]);
+              channelLabel, e.message]);
         });
       }, (e:Error) => {
         log.error('failed to open channel for new SOCKS client: %1 ',
@@ -335,6 +335,8 @@ module SocksToRtc {
       }, this.fulfillStopping_);
 
       // Once shutdown has been requested, free resources.
+      // TODO: This promise is only used for logging, so we should probably
+      // remove it, or at least make it private.
       this.onceStopped = this.onceStopping_.then(this.stopResources_);
 
       return this.onceReady;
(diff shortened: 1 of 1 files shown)


trevj:
Logging messages aside, I really like this. I can see now that `Session.onceStopped` exists only to support the logging statement in the parent class. That logging could (probably should, in fact) be moved to the Session, in which case the removal of `onceStopped`would significantly simplify `Session` and its tests.

That's a bit much for one pull request, but I like where this is going. The only downside I can think of is that it will no longer be possible for `SocksToRtc` to call - for example - `close()` on all of its open sessions; that seems questionable, however, since it's unnecessary with a properly functioning peerconnection object. A little more importantly, we lose some of the count logging, e.g. "5 sessions remaining". Again, not important since we can identify dangling sessions from logs.

For now, can you just make a similar change to `RtcToNet`, i.e. remove its `sessions_` field? Then, :+1:.

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 221 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/248#-JnWXiyqeQqSKWeacnOW)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/8bda30c799bdcc78e9665ae65de1f80d0ac14c7d/src/socks-to-rtc/socks-to-rtc.ts#L221)):</sup>
I think `channelLabel` is more appropriate now.

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 234 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/248#-JnWXuiUOz2uVrdwCEnZ)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/8bda30c799bdcc78e9665ae65de1f80d0ac14c7d/src/socks-to-rtc/socks-to-rtc.ts#L234)):</sup>
Now, `onceStopped` exists only to support this logging statement, suggesting it isn't really useful anymore.

That's interesting, since its removal would significantly simplify `Session`.

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/248)

trevj:
(note to myself: this will obsolete https://github.com/uProxy/uproxy/issues/561)

bemasc:
I looked into doing the same in RtcToNet ... but (1) RtcToNet actually relies on `sessions_` to implement the snapshot behavior, and luckily (2) it doesn't use `sessions_` to prevent colliding sessions from starting, so the only parallel impact should be slightly skewed snapshots, not failing connections.

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 221 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/248#-JnWXiyqeQqSKWeacnOW)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/8bda30c799bdcc78e9665ae65de1f80d0ac14c7d/src/socks-to-rtc/socks-to-rtc.ts#L221)):</sup>
Done.

---

<sup>**[src/socks-to-rtc/socks-to-rtc.ts, line 234 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/248#-JnWXuiUOz2uVrdwCEnZ)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/8bda30c799bdcc78e9665ae65de1f80d0ac14c7d/src/socks-to-rtc/socks-to-rtc.ts#L234)):</sup>
I added a TODO

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/248)

trevj:
I forgot about that. Seems like we will want to re-design how the snapshots are done. We can do that at some point in the future.

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/248)

PR #249, opened 2015-04-22
move to heartbeats-based uproxy-lib

trevj:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/249)


trevj:
Addresses:
https://github.com/uProxy/uproxy/issues/1358

PR #250, opened 2015-04-23
fix paths to loggingprovider in firefox samples

trevj:
Gives us timestamps, and other custom goodness, in the logs.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/250)


bemasc:
:+1:

PR #252, opened 2015-04-24
Implement load tests

bemasc:
These tests are huge, slow, and flaky, but test important proxying functions.

This pull request has previously been reviewed at https://github.com/uProxy/uproxy-networking/pull/230 , which was a pull request to `dev-requirify`.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/252)


bemasc:
Open question: it's been suggested that these tests could serve as some sort of low-precision benchmark.  I don't know of a way to report execution times that would make this useful.

bemasc:
@trevj PTAL
@mollyling do you know of a good way to record/report benchmark scores?

src/integration-tests/socks-echo/slow.core-env.spec.ts
```diff
@@ -2,27 +2,119 @@
...
+
   // The default TCP SYN timeout is two minutes, so to be safe we
   // set a test timeout of four minutes.
   jasmine.DEFAULT_TIMEOUT_INTERVAL = 240000;
 
-  it('attempt to connect to a nonexistent IP address', (done) => {
-    getTestModule().then((testModule:ProxyIntegrationTester) => {
-      // 192.0.2.0/24 is a reserved IP address range.
-      return testModule.connect(80, '192.0.2.111');
+  it('download load test', (done) => {
```
trevj, line 41: one line description would be useful here.

src/integration-tests/socks-echo/slow.core-env.spec.ts
```diff
@@ -2,27 +2,119 @@
...
+        });
+        return testModule.sendData(connectionId, testBlock).then(() => {
+          return result;
+        });
+      });
+      Promise.all(completions).then(done);
+    });
+  });
+
+  it('upload load test', (done) => {
```
trevj, line 78: one line description would be useful here.

src/integration-tests/socks-echo/slow.core-env.spec.ts
bemasc, line 41: Done

src/integration-tests/socks-echo/slow.core-env.spec.ts
bemasc, line 78: Done

trevj:
Nice. Thanks for tidying this up!

Using these a kind of benchmark is an interesting idea; right now, I don't see how they can give us much more than `grunt benchmark`, with the obvious benefits that they:
1. are written using JavaScript code
2. run inside Jasmine, which has good support from various continuous build systems

Tracking performance numbers over time is unfortunately very hard unless you keep the same hardware forever, regardless of how the benchmarks run. The idea that springs to mind is to mind is require that the benchmark numbers on the developer's workstation don't fall by less than ~10% (or something) for each submit.

:+1:

mollyling:
I'd log it to a file controlled by git, with a hostname, timestamp, etc.  Extract the right console.log data by having grunt-jasmine-chromeapp launch chrome with [the right options](https://www.npmjs.com/package/grunt-jasmine-chromeapp) to save logs to a file that you can grep for the data that you want.

mollyling:
You can get pretty good stats from this, just by logging some info about each run: platform, cpu data (#cores, models, grepped from /proc/cpuinfo), and load level (also available by grepping /proc).

bemasc:
OK, given the complexity of the tracking question, I'll leave that alone for now, and just merge as-is.

PR #253, opened 2015-04-28
Add a DefinitelyTyped rule for circular-json

jpevarnek:
We have the really annoying dependency rule that we need to pull in
typings for any child repo since the files get copied to our third_party
directory instead of living in their build directories that have the
typings specified.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/253)


trevj:
Thanks! Can you bump the patch version, too? Then :+1:.

PR #254, opened 2015-04-28
always destroy socket provider on disconnect, and fix type error

trevj:
A new version of the TypeScript compiler highlighted this type error:
https://github.com/uProxy/uproxy-networking/blob/dev/src/net/tcp.ts#L61

Fixing it, I realised two things:
1. `close` throws an error when the socket is already closed
2. we have only been destroying the provider when we explicitly call `close`

I believe it's more appropriate to destroy the provider once we receive an `onDisconnect` event. This may fix:
https://github.com/uProxy/uproxy/issues/1303

@willscott could you also take a tiny quick look at this?

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/254)

### DIFF
--- a/src/integration-tests/tcp/freedom-module.ts
+++ b/src/integration-tests/tcp/freedom-module.ts
@@ -59,12 +59,12 @@ parentModule.on('shutdown', () => {
     var client = new tcp.Connection({endpoint: endpoint});
     server.connectionsQueue.setSyncHandler((connection:tcp.Connection) => {
       client.onceConnected.then(() => {
-    server.shutdown();
+        server.shutdown();
         return Promise.all<any>([connection.onceClosed, client.onceClosed,
             server.onceShutdown()]);
       })
       .then((values:any) => {
-    parentModule.emit('shutdown');
+        parentModule.emit('shutdown');
       });
     });
   });
--- a/src/integration-tests/tcp/freedom-module.ts
+++ b/src/integration-tests/tcp/freedom-module.ts
@@ -59,12 +59,12 @@ parentModule.on('shutdown', () => {
     var client = new tcp.Connection({endpoint: endpoint});
     server.connectionsQueue.setSyncHandler((connection:tcp.Connection) => {
       client.onceConnected.then(() => {
-        server.shutdown();
+    server.shutdown();
         return Promise.all<any>([connection.onceClosed, client.onceClosed,
             server.onceShutdown()]);
       })
       .then((values:any) => {
-        parentModule.emit('shutdown');
+    parentModule.emit('shutdown');
       });
     });
   });
--- a/src/net/tcp.ts
+++ b/src/net/tcp.ts
@@ -6,6 +6,7 @@
 import logging = require('../../../third_party/uproxy-lib/logging/logging');
 import handler = require('../../../third_party/uproxy-lib/handler/queue');
 import net = require('./net.types');
+import counter = require('./counter');
 
 var log :logging.Log = new logging.Log('tcp');
 
@@ -54,33 +55,6 @@ function destroyFreedomSocket_(socket:freedom_TcpSocket.Socket) : void {
   freedom['core.tcpsocket'].close(socket);
 }
 
-// Sandwiches a call to an asynchronous function between two other
-// (synchronous) functions. Intended for tracking the number of calls
-// inflight to a freedom provider, which may fail to resolve if the
-// provider's communication channel is destroyed while the call is
-// pending (this can happen even on a normal call to the provider's
-// close method).
-// Throws synchronously if the before function raises an error; the
-// returned promise rejects if the after function raises an error.
-function wrap<T>(
-    before:() => void,
-    after:() => void,
-    f:() => Promise<T>) : Promise<T> {
-  try {
-    before();
-  } catch (e) {
-    after();
-    throw e;
-  }
-  return f().then((result:T) => {
-    after();
-    return result;
-  }, (e:Error) => {
-    after();
-    throw e;
-  });
-}
-
 // Promise and handler queue-based TCP server with freedomjs sockets.
 // TODO: protection against multiple calls to methods such as listen
 export class Server {
@@ -91,6 +65,9 @@ export class Server {
   private id_ :string;
 
   private socket_ :freedom_TcpSocket.Socket;
+
+  // Tracks calls to the socket, for safe destruction.
+  private counter_ :counter.Counter;
 
   // Active connections to the server.
   // TODO: index by connectionId rather than socketID
@@ -113,9 +90,6 @@ export class Server {
     this.fulfillShutdown_ = F;
   });
 
-  // Number of pending (asynchronous) requests to the socket.
-  private numInflightSocketOps_= 0;
-
   constructor(private endpoint_ :net.Endpoint,
       private maxConnections_ :number = DEFAULT_MAX_CONNECTIONS) {
     this.id_ = 'S' + (Server.numCreations_++);
@@ -127,15 +101,21 @@ export class Server {
     this.socket_ = freedom['core.tcpsocket']();
     this.socket_.on('onConnection', this.onConnectionHandler_);
     this.socket_.on('onDisconnect', this.onDisconnectHandler_);
+
+    this.counter_ = new counter.Counter(destroyFreedomSocket_.bind(
+        undefined, this.socket_));
+    this.counter_.onceDestroyed().then(() => {
+      log.debug('%1: closed socket channel', this.id_);
+    }, (e:Error) => {
+      log.error('%1: error closing socket channel: %2', this.id_, e.message);
+    });
   }
 
   // Invoked when the socket terminates.
   private onDisconnectHandler_ = (info:freedom_TcpSocket.DisconnectInfo) : void => {
     log.debug('%1: disconnected: %2', this.id_, JSON.stringify(info));
 
-    // Call this without incrementing this.inflight_ to drive the counter
-    // below zero, which is the termination condition.
-    this.postSocketOp_();
+    this.counter_.discard();
 
     if (info.errcode === 'SUCCESS') {
       this.fulfillShutdown_(SocketCloseKind.WE_CLOSED_IT);
@@ -145,30 +125,10 @@ export class Server {
     }
   }
 
-  private preSocketOp_ = () : void => {
-    this.numInflightSocketOps_++;
-  }
-
-  // Closes the socket instance's channel if there are no more
-  // inflight calls to the socket.
-  // See #destroyFreedomSocket_.
-  private postSocketOp_ = () : void => {
-    this.numInflightSocketOps_--;
-    if (this.numInflightSocketOps_ < 0) {
-      try {
-        destroyFreedomSocket_(this.socket_);
-        log.debug('%1: closed socket channel', this.id_);
-      } catch (e) {
-        log.error('%1: error closing socket channel: %2',
-            this.id_, e.message);
-      }
-    }
-  }
-
   // Listens for connections, returning onceListening.
   // Should only be called once.
   public listen = () : Promise<net.Endpoint> => {
-    wrap(this.preSocketOp_, this.postSocketOp_, () => {
+    counter.wrap(this.counter_, () => {
       return this.socket_.listen(this.endpoint_.address,
           this.endpoint_.port).then(() => {
         return this.socket_.getInfo();
@@ -243,7 +203,7 @@ export class Server {
   public stopListening = () : Promise<void> => {
     log.debug('%1: closing socket, no new connections will be accepted',
         this.id_);
-    return wrap(this.preSocketOp_, this.postSocketOp_, this.socket_.close);
+    return counter.wrap(this.counter_, this.socket_.close);
   }
 
   // Closes all active connections.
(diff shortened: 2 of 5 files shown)


src/integration-tests/tcp/freedom-module.ts
```diff
@@ -59,12 +59,12 @@ parentModule.on('shutdown', () => {
     var client = new tcp.Connection({endpoint: endpoint});
     server.connectionsQueue.setSyncHandler((connection:tcp.Connection) => {
       client.onceConnected.then(() => {
-        server.shutdown();
+    server.shutdown();
```
bemasc: This spacing seems ... worse.

src/net/tcp.ts
```diff
@@ -46,20 +46,15 @@ export function endpointOfSocketInfo(info:freedom_TcpSocket.SocketInfo)
...
-  };
-  return socket.close().then(destroy, (e:Error) => {
-    destroy();
-    return e;
-  });
+// Closes a socket's freedomjs interface object, i.e.:
+//   freedom['core.tcpsocket']().close
+//
+// This is different from:
+//   freedom['core.tcpsocket'].close
```
bemasc, line 52: Do you mean freedomTcpSocket.close()?
EDIT: I mean, are these two cases backwards?

src/net/tcp.ts
```diff
@@ -116,6 +111,18 @@ export class Server {
       // TODO: investigate which other values occur
       this.fulfillShutdown_(SocketCloseKind.UNKOWN);
     }
+
+    // Since there may be other calls to this provider in progress which will
+    // fail to resolve if we destroy the communication channel now, postpone
+    // doing so until the next iteration of the event loop.
+    setTimeout(() => {
```
bemasc: This seems incredibly fragile.  Is there a reason that this is actually safe?  Could we structure this to wait for a promise so that it's actually deterministic?

src/integration-tests/tcp/freedom-module.ts
trevj: Sorry, I had my diff tool ignoring leading whitespace for some strange reason...

src/net/tcp.ts
trevj, line 52: Oh god, you're right -- and I looked extra hard at that. I shortened this, hopefully it's clearer now (and a lot shorter).

src/net/tcp.ts
trevj: You're right. Probably happenstance that this works right now. I couldn't think of a simple general solution...let me think about it some more.

trevj:
Thanks for looking at this!

src/net/tcp.ts
trevj: OK, it was kinda fun to try to come up with a non-horrible solution...I'm not sure I have, although the integration tests pass now.

This of course is without changing the guarantee on when `onDisconnect` fires, which we might also consider.

trevj:
PTAL!

bemasc:
<sup>**[src/net/tcp.ts, line 67 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoFamcx71nGv2n-fNWp)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/891c934246ca3869621eaae51e619c5c795c196b/src/net/tcp.ts#L67)):</sup>
How about something like

``` typescript
before();
return f().then((val:T) => {
  after()
  return val;
}, (e:Error) => {
  after();
  throw e;
};
```

That would avoid the ungainly `new Promise` construction.

---

<sup>**[src/net/tcp.ts, line 131 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoFp03zvFFBUzHRYSwV)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/891c934246ca3869621eaae51e619c5c795c196b/src/net/tcp.ts#L131)):</sup>
I think this needs a comment to explain that its purpose is to drive the count to -1, which is a special condition indicating termination.

---

<sup>**[src/net/tcp.ts, line 399 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoFpInWhIN1pyIh9QaX)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/891c934246ca3869621eaae51e619c5c795c196b/src/net/tcp.ts#L399)):</sup>
Binding `this` to `undefined` seems pretty scary to me.

---

<sup>**[src/net/tcp.ts, line 428 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoFpO8pcbDLxB-B8Mip)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/891c934246ca3869621eaae51e619c5c795c196b/src/net/tcp.ts#L428)):</sup>
Same here

---

<sup>**[src/net/tcp.ts, line 453 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoFpS2d3-ByOdb-7PBy)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/891c934246ca3869621eaae51e619c5c795c196b/src/net/tcp.ts#L453)):</sup>
Copy and paste makes me sad.  What if the wrapper were its own class, and then `Connection` and `Server` each had one?

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254)

trevj:
Thanks @bemasc, PTAL!

---

<sup>**[src/integration-tests/tcp/freedom-module.ts, line 62 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-Jo1EshnVHq9hprN-VxO-r1-62)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/60549e1cebe945a20d5ab6afd9eda1b318aaf8e7/src/integration-tests/tcp/freedom-module.ts#L62)):</sup>
Done.

---

<sup>**[src/net/tcp.ts, line 51 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-Jo1EshnVHq9hprN-VxP-r1-53)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/60549e1cebe945a20d5ab6afd9eda1b318aaf8e7/src/net/tcp.ts#L51)):</sup>
Done.

---

<sup>**[src/net/tcp.ts, line 67 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoFamcx71nGv2n-fNWp)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/891c934246ca3869621eaae51e619c5c795c196b/src/net/tcp.ts#L67)):</sup>
Right. I did this so any error raised by `after()` would propagate asynchronously but just didn't realised what I had didn't work...changed now. `before()` will just raise eceptions synchronously.

---

<sup>**[src/net/tcp.ts, line 399 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoFpInWhIN1pyIh9QaX)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/891c934246ca3869621eaae51e619c5c795c196b/src/net/tcp.ts#L399)):</sup>
Ah, yes. Probably only works due to how the socket provider is written internally. Fixed, thanks!

---

<sup>**[src/net/tcp.ts, line 453 [r4]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoFpS2d3-ByOdb-7PBy)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/891c934246ca3869621eaae51e619c5c795c196b/src/net/tcp.ts#L453)):</sup>
Right. I've extracted this to a separate file, with a test. It's a little tricky because only standalone functions can be parameterised...but I think this is about as clean as I can get it.

---

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254)

bemasc:
<sup>**[src/net/counter.ts, line 10 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVciQClpEb-JwEnIth)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/counter.ts#L10)):</sup>
I think this would be more attractive as a method of `Counter`.  Then `.before` and `.after` could be private.

---

<sup>**[src/net/counter.ts, line 39 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVdjIOh5idaFFYIUMB)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/counter.ts#L39)):</sup>
I think this pattern needs a comment to explain how it relates to the `this.counter_ < 0` check.  Alternatively, maybe a comment on `counter_` to explain its semantics.

---

<sup>**[src/net/tcp.ts, line 416 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVdZe0UrO2zoHp3GMw)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/tcp.ts#L416)):</sup>
Do you want to wait for this.counter_.onceDestroyed here?

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254)

trevj:
<sup>**[src/net/tcp.ts, line 416 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVdZe0UrO2zoHp3GMw)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/tcp.ts#L416)):</sup>
YES. Excellent point. Done, for server and client sockets. I like it much better now. Thanks.

---

<sup>**[src/net/counter.ts, line 10 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVciQClpEb-JwEnIth)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/counter.ts#L10)):</sup>
So, I agree -- however, I wanted type safety on the return type and TypeScript only supports this on standalone functions.

I don't think there's any way to move this method to `Counter` without making the return type `any`, which I think would be worse.

---

<sup>**[src/net/counter.ts, line 39 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVdjIOh5idaFFYIUMB)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/counter.ts#L39)):</sup>
Done.

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254)

bemasc:
<sup>**[src/net/counter.ts, line 10 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVciQClpEb-JwEnIth)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/counter.ts#L10)):</sup>
Just make `Counter<T>` parametric?

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254)

bemasc:
<sup>**[src/net/counter.ts, line 10 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVciQClpEb-JwEnIth)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/counter.ts#L10)):</sup>
Oh, I see; it's used to wrap several different types.  Hmm...

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254)

trevj:
<sup>**[src/net/counter.ts, line 10 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVciQClpEb-JwEnIth)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/counter.ts#L10)):</sup>
Right. It works great but only on a standalone function.

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254)

bemasc:
<sup>**[src/net/counter.ts, line 10 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVciQClpEb-JwEnIth)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/counter.ts#L10)):</sup>
I just tried this locally.  I think you can do

``` typescript
public wrap = <T>(f:() => Promise<T>) : Promise<T> => {
```

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254)

trevj:
<sup>**[src/net/counter.ts, line 10 [r6]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254#-JoVciQClpEb-JwEnIth)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/6669641b0dde507e3afce6853d84e193cbd5fca4/src/net/counter.ts#L10)):</sup>
Oh! You know, I tried something very similar and couldn't get the magic incantation right. Many thanks, fixed.

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254)

bemasc:
:+1: 

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/254)

PR #256, opened 2015-05-05
workaround for core.udpsocket.getInfo address reporting issue

trevj:
Fixes:
https://github.com/uProxy/uproxy/issues/1263

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/256)


bemasc:
:+1:

PR #257, opened 2015-05-05
defensive programming: onDisconnect should check the supplied value

trevj:
Just a little thing arising from:
https://github.com/freedomjs/freedom-for-firefox/issues/63

Still on the fence about how defensive we want to be...in this case, it's a clear bug in the provider but easy to workaround.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/uproxy-networking/257)

### DIFF
--- a/src/net/tcp.ts
+++ b/src/net/tcp.ts
@@ -108,13 +108,17 @@ export class Server {
 
   // Invoked when the socket terminates.
   private onDisconnectHandler_ = (info:freedom_TcpSocket.DisconnectInfo) : void => {
-    log.debug('%1: disconnected: %2', this.id_, JSON.stringify(info));
+    if (info) {
+      log.debug('%1: onDisconnect: %2', this.id_, info);
+    } else {
+      log.warn('%1: onDisconnect without info', this.id_);
+    }
 
     this.counter_.discard();
 
     this.counter_.onceDestroyed().then(() => {
       log.debug('%1: closed socket channel', this.id_);
-      if (info.errcode === 'SUCCESS') {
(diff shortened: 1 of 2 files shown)


src/net/tcp.ts
```diff
@@ -108,13 +108,17 @@ export class Server {
 
   // Invoked when the socket terminates.
   private onDisconnectHandler_ = (info:freedom_TcpSocket.DisconnectInfo) : void => {
-    log.debug('%1: disconnected: %2', this.id_, JSON.stringify(info));
+    if (info) {
```
bemasc, line 111: Please add a comment linking to the bug.  I think a TODO to remove this might be appropriate; it seems to me that freedom_TcpSocket's API ought to guarantee that `info` is non-null.

trevj:
<sup>**[src/net/tcp.ts, line 111 [r1]](https://reviewable.io:443/reviews/uproxy/uproxy-networking/257#-Jo_TBdzfG-xWnwSqpIm-r1-111)** ([raw file](https://github.com/uproxy/uproxy-networking/blob/5fed8fe66f18910f1840a847a9c45cf232ff76d4/src/net/tcp.ts#L111)):</sup>
Done.

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/uproxy-networking/257)

bemasc:
:+1:, although a unit test would be nice.

trevj:
Right. I tried that just now and the `Connection` enums are defined in a slightly weird way -- in a TypeScript module. I dunno why they're done that way so I think I'd prefer do this as part of a general `Connection` cleanup, following on from my recent server socket cleanup.

